PR #7, opened 2023-04-11
custom color theming

choe220:
Added a ThemeNotifierProvider to handle brightness state and color source.
Added a color picker to the RequestEditorPaneHome to select a custom color theme.
Updated the blend mode of the sendingIndicator on SendingWidget to better match both color theme and brightness.


choe220:
Fixed a bug that wouldn't allow the theme to persist by adding a check to the removeUnused method inside of HiveHandler

animator:
Hi @choe220 currently we are focussing only on dark mode, so putting this PR on a hold.

PR #34, opened 2023-05-22
Update CONTRIBUTING.md

mmjsmohit:
I added instructions to avoid the invalid Dart Package name error occurring when the project is run for the first time after cloning. It caused the invalid name "api-dash," with which the machine creates the repository by default.


animator:
Hi @mmjsmohit, Thanks for adding this note.

Issue #33, opened 2023-05-21
Mobile Android/ iOS Support


Kop3sh:
**Tell us about the task you want to perform and are unable to do so because the feature is not available**
I would like to add support for Mobile Platforms i.e. Android, iOS this would allow to test APIs on-the-go. It might seem excessive and unnecessary but this kind of agility that enables everyone to test stuff on-the-go, having the same familiar UX and set of features providing continuity, with the added benefit of having a shared open-source codebase powered by flutter is what could potentially set this app apart from other competition.

**Describe the solution/feature you'd like us to add**
My proposal is building a preliminary mobile version of the app with a simplified UI/UX that can empower users to test APIs and endpoints in the simplest way possible on-the-go. This could potentially be utilized in adding future features like integrating a cross-platform sync; providing continuity for signed-up users.

**Approach**
This is obviously a fairly big feature to be added, for that my approach would be to start
1. Designing an interface with the bare minimum of functional features coupled with the same beautiful design, (no changes, just adapting for smaller screens)
2. Iteratively develop/ adapt code for the new platforms (starting with Android because it's usually more accessible)
3. Test and add more features while maintaining this simple design language.

animator:
Hi @Kop3sh,

Thanks for opening this issue and taking a lead on adding Mobile Android/ iOS Support.
One of the reasons we chose Flutter was to ship API Dash across all platforms one day. Adding Android/ iOS support will definitely be an ambitious project and will require a lot of work, but your step-by-step approach is definitely the way to go about it. Before you start coding Step 1, do share the UI mockups for smaller screens in this issue thread, so that we can discuss what would be the best way to adapt the UI for mobile. Also, do join our Discord server and ping **#foss** channel in case you need any help.

Kop3sh:
Hi @animator 

Thank you...

I'll get to it then.
It would be great if you could share with me existing UI mockups/ components/ themes...  so that I repurpose them and follow the same design rules without reinventing the wheel and wasting precious time ?

animator:
We do not have separate mockups currently as we are using Material 3 as the base. And any tweaks made and all reusable components of API Dash are present in `lib/consts.dart` & `lib/widgets/`.

animator:
Also @Kop3sh, by mockups, I just meant simple rough drawings to show what elements (button, tabs, fields, etc.) will go where for smaller screens. There is no need of very elaborate mockups with full themes. Just simple drawings to discuss on the layout.

Kop3sh:
[...]

PR #70, opened 2023-07-01
Initial Mobile Support

Kop3sh:
### Added initial mobile support according to the issue #33 latest discussions
N.B. I moved home, requests and settings routes within the side drawer instead of the main scaffold body for better UX in my opinion

- added platform specific consts for mobile, iOS and android
- created a preliminary layout for mobile devices by creating a new dynamic MobileDashboard widget.
- added in widget platform specific checks to display widgets accordingly.
- added basic MaterialPageRoutes for home, requests and settings within the side drawer
- utilized mostly the same existing widgets. (servers as a starting point 'PoC' that can be refined later down the road for adaptive layout and visual density)

welcoming notes, reviews and architectural/ code organization guidance


animator:
Thanks @Kop3sh!

I ran the PR in the simulator and actually this was looking very neat. I think you should definitely proceed forward to add more features.

A few pointers from project structure point of view.
- `lib/widgets` was created to contain only pure flutter widgets (Stateful/Stateless) that have no dependency on riverpod. So `mobile_dashboard.dart` should reside beside `dashboard.dart` inside `lib/screens/`.
- Now, this is where things get tricky and we can end up having a mess pretty soon if mobile specific screen files are placed alongside desktop versions. Since we want to reuse as much code as possible, let us create `lib/screens/mobile/` folder and place mobile specific screens inside it like `mobile_dashboard.dart` (renamed as `dashboard.dart` inside `lib/screens/mobile/`). Now, we can do relative imports and reuse as much code as we can and also segregate any code we specifically wrote for the mobile version.

Kop3sh:
totally agree with your approach...

made the required architectural changes as requested, and left the MobileDashboard widget name as is (to prevent confusion with conventional Dashboard)

lib/app.dart
```diff
@@ -1,9 +1,12 @@
+import 'package:apidash/screens/home_page/collection_pane.dart';
+import 'package:apidash/screens/mobile/mobile.dart';
```
animator, line 2: `screens/screens.dart` is already imported ...
You should add `export 'screens/mobile/mobile.dart'` to `lib/screens/screens.dart`. Also, if you need to expose `CollectionPane` to the app.dart, you can add `home_page/collection_pane.dart` to the `lib/screens/screens.dart`. Anything exposed outside the screens folder should be through that file.

lib/screens/mobile/dashboard.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:apidash/screens/home_page/collection_pane.dart';
+import 'package:apidash/screens/intro_page.dart';
+import 'package:apidash/screens/settings_page.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
```
animator, line 5: As a convention we follow this sequence:

- import 'package:flutter/material.dart';
- Then we import external packages like riverpod
- Finally we import the files of the current package.

animator: Any file that reside in the same folder (here `screens`) is also imported relatively.
```
import '../intro_page.dart';
import '../settings_page.dart';
import '../home_page/collection_pane.dart';
```

Kop3sh: thanks for clearing that out, I've just finished cleaning as up the imports as described in all of the comments, and it will be pushed in the next commit...
I'm curious though, is going with relative imports for the second comment more favorable to prevent circular dependencies (realization/ guess) or is it just a convention... ?
because, I've tried both ways and it still works, searched further and found out that dart's multipass compiler permits circular imports as well.

animator: No the relative import is not to prevent circular dependencies (we dont have any) .. just a convention to quickly gauge that there are widgets being used whose source is nearby (screens ,widgets ,etc. wherever they reside).

Kop3sh: Got it

Kop3sh:
I cleaned up the imports and finished the requested changes...
Is there any other pointers/ reviews

lib/screens/screens.dart
```diff
@@ -1 +1,3 @@
-export "dashboard.dart";
+export 'dashboard.dart';
+export 'mobile/mobile.dart';
+export 'home_page/collection_pane.dart';
```
animator, line 3: Is `export 'home_page/collection_pane.dart';` required here ...? as no one is using it outside the screens folder.

animator:
Just one slight comment I posted above ... apart from that you are good to go and add the rest of the functionality. 
Looking forward to the next set of commits.
Also, kindly make sure that you sync your fork & branch time to time to pull any changes and work on the latest head.

lib/screens/screens.dart
Kop3sh: actually, it is used in the /lib/app.dart, within the MobileDashboard widget as scaffoldBody...

Kop3sh: actually, it is used in the /lib/app.dart, within the MobileDashboard widget as scaffoldBody...

Kop3sh:
yep, thanks for the reminder...
I just synced the fork as pointed out.

lib/screens/screens.dart
animator: Cool.. Then it makes sense.

animator:
Hi @Kop3sh .. Just wanted to check with you how things are going on with the mobile UI.
Cheers!

Kop3sh:
> Hi @Kop3sh .. Just wanted to check with you how things are going on with the mobile UI. Cheers!

Sorry for being late, going through personal matters.
but yes, everything is good regarding the mobile UI, I'll get back to you with the next set of commits ASAP.

animator:
No worries @Kop3sh .. Please take care!

Issue #27, opened 2023-05-20
Autocomplete/Autosuggestions while entering headers.


aqsasayyed:
#### Tell us about the task you want to perform and are unable to do so because the feature is not available
While typing headers there are no suggestions. Adding suggestions will improve the UX.

#### Describe the solution/feature you'd like us to add
Solution -
While typing inside headers textfield let's give users auto suggestion.

Existing -
<img width="758" alt="image" src="https://github.com/foss42/api-dash/assets/84732662/cd12c7fb-dd68-489f-8503-f7764553579b">

Proposed - 
<img width="844" alt="image" src="https://github.com/foss42/api-dash/assets/84732662/3a7ea4b3-cc74-4e42-95fc-145f83ccf6cc">
#### Any other feedback you would like to provide regarding the site
N/A

#### Are you working on this issue/Would you link to work on this issue?
Yes

ashitaprasad:
Hi @aqsasayyed , 
Firstly, Join our [discord server](https://bit.ly/heyfoss) so that you can message on the **#foss** channel whenever you encounter any issue. That is also our official channel where we have all discussions related to API Dash. You must also follow the [Contribution Guidelines](https://github.com/foss42/api-dash/blob/main/CONTRIBUTING.md). 

This is a nice feature 🙂 to have in API Dash. You can go ahead and implement it and raise a PR for code review.

MihirCharan42:
Hi i would like to work on this issue as well, could you please assign this to me? I am GSSoC'2023 participant

ashitaprasad:
Hi @MihirCharan42, 
As this new feature suggestion was made by @aqsasayyed, she is currently assigned to & working on this issue.

aqsasayyed:
@ashitaprasad I'm working on this enhancement. I'll be creating a PR once done.

Thanks

animator:
@aqsasayyed any updates on adding this enhancement?

aqsasayyed:
Hi @animator actually my exams are going so give me some time I'll close this in this week only. I have completed it 50%.

animator:
No worries @aqsasayyed .. I was not aware of the exams. best of luck!

PR #83, opened 2023-10-20
Add feature header suggestions

DenserMeerkat:
- Used [flutter_typeahed](https://pub.dev/packages/flutter_typeahead) to generate suggestions while typing headers.
- Added a few headers for now (can be expanded in future) as key : value pairs with key as the header and its use as the value, which can be used in future for showing tooltips or popovers.
- This PR closes https://github.com/foss42/apidash/issues/27.


animator:
Hi @DenserMeerkat ,
Thanks for adding this feature .. the popup goes very well with the theme and it was sleek and very useful.
Behold the merge! 🎉

PR #86, opened 2023-11-16
Add feature header checkbox

DenserMeerkat:
Added the checkbox functionality by introducing a `enabled` attribute to the NameValueModel.

### DIFF
--- a/lib/models/name_value_model.dart
+++ b/lib/models/name_value_model.dart
@@ -8,6 +8,7 @@ part 'name_value_model.g.dart';
 @freezed
 class NameValueModel with _$NameValueModel {
   const factory NameValueModel({
+    @Default(true) bool enabled,
     required String name,
     required dynamic value,
   }) = _NameValueModel;
--- a/lib/models/request_model.dart
+++ b/lib/models/request_model.dart
@@ -132,18 +132,18 @@ class RequestModel {
     if (requestHeaders != null) {
       try {
         requestHeadersList =
-            mapToRows(Map<String, String>.from(requestHeaders));
+            mapToRows1(Map<String, List<dynamic>>.from(requestHeaders));
       } catch (e) {
         requestHeadersList =
-            mapToRows1(Map<String, List<dynamic>>.from(requestHeaders));
+            mapToRows(Map<String, String>.from(requestHeaders));
       }
     }
     if (requestParams != null) {
       try {
-        requestParamsList = mapToRows(Map<String, String>.from(requestParams));
-      } catch (e) {
         requestParamsList =
             mapToRows1(Map<String, List<dynamic>>.from(requestParams));
+      } catch (e) {
+        requestParamsList = mapToRows(Map<String, String>.from(requestParams));
       }
     }
 
--- a/lib/models/request_model.dart
+++ b/lib/models/request_model.dart
@@ -40,7 +40,9 @@ class RequestModel {
   Map<String, String> get headersMap =>
       rowsToMap(requestHeaders?.where((element) => element.enabled).toList()) ??
       {};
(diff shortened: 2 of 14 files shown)


ashitaprasad:
Hi @DenserMeerkat ,
Please add test cases for these changes.
More details - [Link](https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md#how-to-run-tests)

DenserMeerkat:
> Hi @DenserMeerkat , Please add test cases for these changes. More details - [Link](https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md#how-to-run-tests)

👍 Ok, will  add the required tests and make a commit.

DenserMeerkat:
Hi @ashitaprasad, I've fixed the existing tests for the model changes and added tests for the newly introduced conversion methods.

animator:
Hi @DenserMeerkat, Apologies as we got busy with the release. We will review this PR this week.

DenserMeerkat:
✌🏼 No issues.

lib/models/name_value_model.dart
```diff
@@ -8,6 +8,7 @@ part 'name_value_model.g.dart';
 @freezed
 class NameValueModel with _$NameValueModel {
   const factory NameValueModel({
+    @Default(true) bool enabled,
```
animator, line 11: NameValueModel must not be modified as it should contain only key name and corresponding value. The enabled param should be decoupled from this model.

lib/models/request_model.dart
animator, line 1: In the RequestModel, we are storing the headers & params as a list. Here we can have two new lists of indices of checked items. null would mean all checkboxes are selected. This will also ensure that the code is compatible with the past schema & does not break during importing or reading previous version data as these values will be null. It will also simplify the code to a great extent.

lib/utils/convert_utils.dart
```diff
@@ -78,13 +78,43 @@ Map<String, String>? rowsToMap(List<NameValueModel>? kvRows,
   return finalMap;
 }
 
+Map<String, List<dynamic>>? rowsToMap1(List<NameValueModel>? kvRows,
```
animator, line 81: rowsToMap1 & mapToRows1 naming is improper, and can be removed as NameValueModel will not be changed.

lib/widgets/checkbox.dart
```diff
@@ -0,0 +1,58 @@
+import 'package:flutter/material.dart';
+
+class HeaderCheckBox extends StatefulWidget {
```
animator, line 3: Checkbox color is not following the material design 3 colour. In dark mode it fails to adapt. 
It should also be a stateless widget whose enabled disabled value is provided by the model. and when the user taps on it .. it triggers change of the list of enabled indices in request model which in turn provides new value for it.

test/models/request_model_test.dart
```diff
@@ -103,8 +103,8 @@ void main() {
     "name": 'foss42 api',
     'description': '',
     "requestHeaders": {
-      'content-length': '18',
-      'content-type': 'application/json; charset=utf-8'
+      'content-length': [true, '18'],
```
animator, line 106: request headers cannot be a map of key:list that is why change in request model is required and not in NameValueModel.

animator (commented):
Please take a look at my comments and feel free to reach out in case of any difficulty. 
Thanks!

DenserMeerkat:
> Please take a look at my comments and feel free to reach out in case of any difficulty. Thanks!

@animator Thanks for the review and corrections.
Will open a new PR which follows the mentioned backward compatible approach to handle checked items to implement the checkbox functionality.

animator:
Sure @DenserMeerkat 👍
Feel free to ping on discord channel if you need any help.

PR #94, opened 2023-11-30
[Windows]: Custom Title Bar

### LINKED ISSUE #89: [Windows] Hide/custom title bar like in macOS
Increases space for app by roughly 30 pixels vertically

Add custom title bar, since you are already using `window_manager` anyway, why not also Windows?

**Any other feedback you would like to provide regarding the app**
What I think the app would roughly look like
temp_

damywise:
Fix #89

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -11,13 +11,14 @@ void main() async {
   await openBoxes();
   if (kIsLinux) {
     await setupInitialWindow();
-  } else {
-    var win = getInitialSize();
-    await setupWindow(sz: win.$1, off: win.$2);
   }
   runApp(
     ProviderScope(
       child: kIsLinux ? const DashApp() : const App(),
     ),
   );
+  if (!kIsLinux) {
+    var win = getInitialSize();
+    await setupWindow(sz: win.$1, off: win.$2);
+  }
 }
--- a/lib/screens/dashboard.dart
+++ b/lib/screens/dashboard.dart
@@ -18,6 +18,7 @@ class _DashboardState extends ConsumerState<Dashboard> {
   Widget build(BuildContext context) {
     final railIdx = ref.watch(navRailIndexStateProvider);
     return Scaffold(
+      backgroundColor: Colors.transparent,
       body: SafeArea(
         child: Row(
           children: <Widget>[
--- a/lib/screens/dashboard.dart
+++ b/lib/screens/dashboard.dart
@@ -18,7 +18,7 @@ class _DashboardState extends ConsumerState<Dashboard> {
   Widget build(BuildContext context) {
     final railIdx = ref.watch(navRailIndexStateProvider);
     return Scaffold(
-      backgroundColor: Colors.transparent,
+      backgroundColor: kIsWindows ? Colors.transparent : null,
       body: SafeArea(
         child: Row(
           children: <Widget>[
--- a/lib/services/window_services.dart
+++ b/lib/services/window_services.dart
@@ -41,6 +41,7 @@ Future<void> setupWindow({Size? sz, Offset? off, bool center = false}) async {
   if (kIsDesktop) {
     if (kIsWindows) {
       Window.initialize();
+      Window.hideWindowControls();
     }
     double width = kMinInitialWindowWidth, height = kMinInitialWindowHeight;
     if (sz == null) {
@@ -76,7 +77,6 @@ Future<void> setupWindow({Size? sz, Offset? off, bool center = false}) async {
       await windowManager.focus();
     });
     if (kIsWindows) {
-      Window.hideWindowControls();
       Window.setEffect(effect: WindowEffect.mica);
     }
   }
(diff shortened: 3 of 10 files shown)


pubspec.yaml
```diff
@@ -30,7 +30,10 @@ dependencies:
   lottie: ^2.6.0
   mime_dart: ^3.0.0
   path_provider: ^2.1.0
-  window_manager: ^0.3.5
+  window_manager: 
+    git:
+      url: https://github.com/damywise/window_manager
+      branch: acrylic
```
damywise, line 36: This is a PR I made, see https://github.com/leanflutter/window_manager/pull/374

lib/widgets/window_caption.dart
damywise, line 1: Custom WindowCaption, with removed double click = maximize, because of all the buttons and fields at the top.

lib/services/window_services.dart
```diff
@@ -62,14 +66,20 @@ Future<void> setupWindow({Size? sz, Offset? off, bool center = false}) async {
...
+    windowManager.setClosable(false);
     if (off != null) {
       await windowManager.setPosition(off);
     }
     await windowManager.waitUntilReadyToShow(windowOptions, () async {
       await windowManager.show();
       await windowManager.focus();
     });
+    if (kIsWindows) {
+      Window.setEffect(effect: WindowEffect.mica);
```
damywise, line 82: idk how to get whether the app is dark mode or light mode, so the effect can sync to the app theme

animator: We can get the initial mode here, but any change of mode triggered by user might have to be linked with this function call.

pubspec.yaml
animator: It would be great if you can get this PR merged.

lib/screens/dashboard.dart
```diff
@@ -18,6 +18,7 @@ class _DashboardState extends ConsumerState<Dashboard> {
   Widget build(BuildContext context) {
     final railIdx = ref.watch(navRailIndexStateProvider);
     return Scaffold(
+      backgroundColor: Colors.transparent,
```
animator, line 21: What happens on macos & linux if we set backgroundColor: Colors.transparent
Will need to investigate.

animator:
Thanks for the PR @damywise .. I added some comments.

pubspec.yaml
damywise: Unfortunately I can't get a hold of the author so I can't guarantee you on that.

lib/main.dart
```diff
@@ -11,13 +11,14 @@ void main() async {
...
-  } else {
-    var win = getInitialSize();
-    await setupWindow(sz: win.$1, off: win.$2);
   }
   runApp(
     ProviderScope(
       child: kIsLinux ? const DashApp() : const App(),
     ),
   );
+  if (!kIsLinux) {
```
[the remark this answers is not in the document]
damywise: Can you elaborate on the issue?

For me, I put the window management stuff after the `runApp` because:
1. The features I added introduces an issue where the app would have a white frame for a bit at startup. See the video below.
2. It does not matter as I believe nothing regarding the app itself, not the window, is modified at that step.

Video when putting the line before `runApp`"

https://github.com/foss42/apidash/assets/25608913/da746e47-acf4-46cc-8a96-c7c87844d25c

Video when putting the line after `runApp`"

https://github.com/foss42/apidash/assets/25608913/7bd3a53c-b522-4488-b863-379d848e0f26

lib/services/window_services.dart
damywise: Sorry I don't understand, can you elaborate?

lib/widgets/window_caption.dart
damywise: Might not be desirable for users that would like to maximize through double click

pubspec.yaml
animator: I think a reason it might not be getting merged is the dependency on `flutter_acrylic`

lib/main.dart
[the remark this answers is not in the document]
animator: okay got it.

pubspec.yaml
damywise: That was on the example app which I believe isn't related to the package.
The change on the plugin itself is minimal.
Well I can't know for sure since I haven't been contacted by any means of communications.

animator:
Is it possible to add just the minimize, maximize & close buttons & use the current window_manager package on pubdev without having any dependency on `flutter_acrylic` package. We are planning to stick to the current theme & not use the acrylic (glassy) theme.

damywise:
Yeah that's possible. I just opened a PR on what I experimented on in the issue.

animator:
Yeah not a problem .. it is actually easier to compare changes via PR hence I requested you to open one.
You can open a new PR and close this one.

pubspec.yaml
animator: Oh .. it is definitely sad to work on something and then get no response at all.

PR #95, opened 2023-12-01
feat(windows): custom title bar

### LINKED ISSUE #89: [Windows] Hide/custom title bar like in macOS

damywise:
Follow up from #94 
Fix #89

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -1,10 +1,16 @@
+import 'package:apidash/widgets/window_caption.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:window_manager/window_manager.dart';
+import 'package:window_manager/window_manager.dart' hide WindowCaption;
 import 'providers/providers.dart';
 import 'screens/screens.dart';
 import 'consts.dart'
-    show kIsLinux, kFontFamily, kFontFamilyFallback, kColorSchemeSeed;
+    show
+        kIsWindows,
+        kIsLinux,
+        kFontFamily,
+        kFontFamilyFallback,
+        kColorSchemeSeed;
 
 class App extends ConsumerStatefulWidget {
   const App({super.key});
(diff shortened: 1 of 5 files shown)


lib/app.dart
```diff
@@ -63,22 +69,37 @@ class _DashAppState extends ConsumerState<DashApp> {
         ref.watch(settingsProvider.select((value) => value.isDark));
     return MaterialApp(
       debugShowCheckedModeBanner: false,
-      theme: ThemeData(
-        fontFamily: kFontFamily,
-        fontFamilyFallback: kFontFamilyFallback,
-        colorSchemeSeed: kColorSchemeSeed,
-        useMaterial3: true,
-        brightness: Brightness.light,
+      home: Stack(
```
damywise, line 72: What I did here is added the title bar (draggable widget, window controls) above the app so overlays doesn't hinder it.
However, this solution requires another MaterialApp, which hopefully doesn't prove as an issue.

lib/main.dart
```diff
@@ -11,13 +11,14 @@ void main() async {
...
-  } else {
-    var win = getInitialSize();
-    await setupWindow(sz: win.$1, off: win.$2);
   }
   runApp(
     const ProviderScope(
       child: DashApp(),
     ),
   );
+  if (!kIsLinux) {
```
damywise, line 20: As before, setting up window after `runApp` to prevent white frame on app startup

animator:
Hi @damywise,

I will test out the PR and add my comments.

lib/main.dart
animator: Yes .. I verified it and it resolves the white frame issue.

lib/app.dart
animator: For now, I think it will work .. might change it in future when we have a better option.

animator: But, this is leading to an issue in macos layout. So I will look further into it for a fix.

animator:
Comments after review:
Works like a charm on Windows.
leads to an initial layout issue in macos which needs further investigation.

damywise:
Ahh I think I figured out what's wrong with both Windows and macOS.

[This](https://github.com/leanflutter/window_manager/#hidden-at-launch) is not possible in this project specifically because `windows` and `macos` are `.gitignore`d.

Why are they ignored?

damywise:
Sorry that got deleted but basically I said that [this](https://github.com/leanflutter/window_manager/#hidden-at-launch) not being possible to implement is one of the cause.

damywise:
I have tried using it in macos and I don't see any problem, basically.

Also try using profile/release mode, usually in debug mode it's quite worse.

animator:
Thanks @damywise I will take a look at it again.

ashitaprasad:
Hi @damywise 
We are going ahead with the merge. Some fixes were made as window constraints was not working on resize if `setupWindow` was called after `runApp`.
A big thank you for your contribution 🎉  Looking forward to your future contributions to the project. 🙂

damywise:
> window constraints was not working on resize if setupWindow was called after runApp.

Didn't know that was a thing but hey, glad it works now!

(To be honest the only thing that should be after `runApp` is `show()` and `focus()` but for some reason I thought it'd be easier to put the whole `setupWindow` after `runApp`)

ashitaprasad:
> (To be honest the only thing that should be after `runApp` is `show()` and `focus()` but for some reason I thought it'd be easier to put the whole `setupWindow` after `runApp`)

Sure .. I will run a few tests to see if putting `show()` and `focus()` after `runApp` solves the issue.

Issue #20, opened 2023-05-05
Enhance SVG previewer


ashitaprasad:
**Use case**

Currently, the app shows svg as an xml code instead of an image unlike other image formats. 
User should be able to directly view SVG in the response pane when the response is of content type image/svg+xml.

Utkarsh4517:
I want to work on it, can you guide me on this @ashitaprasad

ashitaprasad:
Sure @Utkarsh4517, 
Join our [discord server](https://bit.ly/heyfoss) so that you can message on the **#foss** channel whenever you encounter any issue. That is also our official channel where we have all discussions related to API Dash.
You must also follow the [Contribution Guidelines](https://github.com/foss42/api-dash/blob/main/CONTRIBUTING.md).
Before sending a PR, just let me know when your code is ready for review.

Khushal747:
please assign me this issue under GSSOC'23

animator:
@Khushal747 Please follow the instructions provided by @ashitaprasad [here](https://github.com/foss42/api-dash/issues/20#issuecomment-1555879929).

Khushal747:
ok

mmjsmohit:
Since there hasn't been any activity on this issue since a long time, can I work on this one?

animator:
@mmjsmohit Go ahead man 👍

PR #98, opened 2023-12-15
Add SVG Rendering

mmjsmohit:
This PR aims to solve #20 
Added SVG Preview support to `previewer.dart` using the `flutter_svg` package. Added error messages to handle when the response is an invalid SVG.
The SVG Previewer now displays the SVG Image, XML Code for SVG Image and the Raw Response body.
Planning to add tests for the SVG Previewer sometime soon.


animator:
Thanks for the PR 🚀

Issue #90, opened 2023-11-29
Ability to override content-type for POST request


ashitaprasad:
Currently, Content-Type is set as `application/json; charset=utf-8` for POST requests with JSON body. Some legacy APIs only accept `application/json` as content-type in request headers which leads to error. There should be an option to override it.

jassirr:
Hello, @ashitaprasad,
can i take up this issue!

ashitaprasad:
Sure @jassirr waiting for your PR.
If you have any doubt/query feel free to post it in our discord channel [#foss-apidash](https://bit.ly/fgfossd)

ashitaprasad:
The encoding (`charset`) is compulsorily passed by `http` library so the exact use case described is currently not possible. But, still solving this issue helps pass custom content-type for POST requests.

PR #104, opened 2023-12-30
feat: ability to override content type

nobodydotcode:
Added:

1. Content-Type header
2. Suggestions for Content-Type header values

Improvements that we can do:

1. As of now I have used the same HeaderField widget for content-type header value widget also for auto suggestions. We can write a separate widget.
2. Do we have to restrict this header only for POST requests?

### DIFF
--- a/lib/codegen/dart/http.dart
+++ b/lib/codegen/dart/http.dart
@@ -55,8 +55,11 @@ class DartHttpCodeGen {
       final strContent = CodeExpression(Code('r\'\'\'$body\'\'\''));
       dataExp = declareVar('body', type: refer('String')).assign(strContent);
 
-      composeHeaders.putIfAbsent(HttpHeaders.contentTypeHeader,
+      final hasContentTypeHeader = composeHeaders.keys.any((k) => k.toLowerCase() == HttpHeaders.contentTypeHeader);
+      if (!hasContentTypeHeader) {
+        composeHeaders.putIfAbsent(HttpHeaders.contentTypeHeader,
           () => kContentTypeMap[contentType] ?? '');
+      }
     }
 
     Expression? queryParamExp;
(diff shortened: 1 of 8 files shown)


nobodydotcode:
@ashitaprasad Can you please review this PR whenever you have sometime?

ashitaprasad:
Currently, we add Content-Type header [here](https://github.com/foss42/apidash/blob/3423eb7d8ff44accb75ad633b1c2a2bc9eb83111/lib/services/http_service.dart#L31) for POST request. 
To resolve the issue we need to check if `Content-Type` is already available in the `headers` map. If it is then we will not set it.

nobodydotcode:
@ashitaprasad Done. Added the condition. Am I missing anything else? Also I see that Content type header value suggestions have been removed, I guess that is not required at this stage.

ashitaprasad:
@nobodydotcode The value of [`HttpHeaders.contentTypeHeader`](https://api.dart.dev/stable/3.2.4/dart-io/HttpHeaders/contentTypeHeader-constant.html) is `"content-type"`. What if the user enters `Content-Type` as the header field name. 
The current check you have added will fail.

Also, content type header value suggestions have been removed because the user is already providing it via content type dropdown in the Body. This feature is actually a corner case where the user explicitly wants to specify the content type because the API wants some additional info in it as well like charset.

nobodydotcode:
@ashitaprasad I have corrected the check. And I have added one more commit for codegen. And as a condition I think following would be enough.

`headers.keys.any((k) => k.toLowerCase() == HttpHeaders.contentTypeHeader)`

ashitaprasad:
Thank @nobodydotcode I will review the chnages.

ashitaprasad:
Thanks for the PR @nobodydotcode 🙌
LGTM 🚀
Fixes #90

PR #107, opened 2024-01-01
♻️ added: JSON Beautify 

priyaranjan12345:


priyaranjan12345:
api_dash_beautify
api_dash_beautify1

lib/widgets/request_widgets.dart
```diff
@@ -46,13 +46,14 @@ class _RequestPaneState extends State<RequestPane>
...
     }
     return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
       children: [
         Padding(
           padding: kPh20v10,
           child: SizedBox(
             height: kHeaderHeight,
-            child: Row(
-              mainAxisAlignment: MainAxisAlignment.spaceBetween,
```
ashitaprasad, line 55: Why did you modify Row into Wrap?

lib/widgets/request_widgets.dart
```diff
@@ -65,10 +66,9 @@ class _RequestPaneState extends State<RequestPane>
                         ? Icons.code_off_rounded
                         : Icons.code_rounded,
                   ),
-                  label: SizedBox(
```
ashitaprasad, line 68: Again. What is the rationale behind this change.

lib/widgets/response_widgets.dart
```diff
@@ -407,22 +408,31 @@ class _BodySuccessState extends State<BodySuccess> {
         return Padding(
           padding: kP10,
           child: Column(
+            crossAxisAlignment: CrossAxisAlignment.stretch,
```
ashitaprasad, line 411: Why these changes till line 451?

lib/widgets/raw.dart
```diff
@@ -0,0 +1,59 @@
+import 'dart:convert';
+
+import 'package:apidash/providers/providers.dart';
+import 'package:apidash/widgets/snackbars.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+class Raw extends StatelessWidget {
```
ashitaprasad, line 8: Raw can be any type of data. Currently, you are assuming that it is JSON.

ashitaprasad:
The purpose of `Raw` is to show the raw response. But, some users will find this feature useful so in stead of adding an obstructing button, it can be provided as a switch in the setting where the user can select whether they want to see formatted response or the raw unformatted response for JSON response.

lib/widgets/request_widgets.dart
priyaranjan12345: due to overflow issue I changed row to wrap. Also I have changed `preview` and `raw` button to short button on small screen size
screen overflow

ashitaprasad: We already have min width requirement in place which is enforced in release version, so it is not required.

PR #159, opened 2024-02-23
feat: add codegen for rust(reqwest)

### LINKED ISSUE #150: Add CodeGen for Rust (reqwest)

Add CodeGen for Rust
Library -https://github.com/seanmonstar/reqwest

Tanish2002:
Closes #150


animator:
Taking headers as a string map is not correct.
Either you should do it like params.
Or do like we have done in Kotlin codegen
There are better ways ([example](https://stackoverflow.com/questions/47911513/how-do-i-set-the-request-headers-using-reqwest)).

Tanish2002:
got it!

Tanish2002:
@animator 
**Problem**: The current codegen for JSON type POST requests in Rust(reqwest) introduces a dependency on the `serde_json` library. While functional, I'm exploring whether to remove this dependency.

# Original Approach (with `serde_json`)
```rs
fn main() -> Result<(), Box<dyn std::error::Error>> {
   let client = reqwest::blocking::Client::new();
   let url = "https://api.foss42.com/case/lower";

   let payload = serde_json::json!({"text": "LOWER THIS"});

   let response = client
       .post(url)
       .json(&payload)
       .send()?;

   println!("Status Code: {}", response.status()); 
   println!("Response Body: {}", response.text()?);

   Ok(())
}
```
- **Pros:**
  - Type safety and validation. `serde_json` ensures the payload is valid JSON before sending
- **Cons:**
  - Introduces an external dependency.

# Alternative Approach (without `serde_json`)
```rs
fn main() -> Result<(), Box<dyn std::error::Error>> {
   let client = reqwest::blocking::Client::new();
   let url = "https://api.foss42.com/case/lower";

   let payload = r#"{"text": "LOWER THIS"}"#;

   let response = client
       .post(url)
       .header("content-type", "application/json")
       .body(payload)
       .send()?;

   println!("Status Code: {}", response.status());
   println!("Response Body: {}", response.text()?);

   Ok(())
}
```
- **Pros:**
  - Removes the serde_json dependency.
- **Cons:**
  - Loses type safety and validation. The server will receive the payload regardless of its JSON validity.
  
# Final Options
- **Option 1 (Remove serde_json):** Simplifies dependencies but sacrifices build time error checking. Which is what most rust developers would want.
- **Option 2 (Keep serde_json):** Enforces well-formed JSON but adds a dependency.


# My thoughts
I think **Option 1** would be a better choice since most rust developers would prefer to use `serde_json` anyways since rust's selling point it finding errors during build time.

animator:
Can you kindly update the test cases to include example Request Models POST 4 to POST 9.

Tanish2002:
@animator done!

ashitaprasad:
LGTM 🚀

Issue #130, opened 2024-02-22
Add CodeGen for Go (http)


ashitaprasad:
**Tell us about the task you want to perform and are unable to do so because the feature is not available**
Add code generator for Go 
Library - https://pkg.go.dev/net/http

CodeHariK:
Can you elaborate

animator:
@CodeHariK To get a fair idea of the API Dash codegen, check out the codes in the `lib/codegen` folder for languages that have been already implemented like python, kotlin etc. Usually implementing a codegen for a new language/library in API Dash requires:

1. Some idea of how jinja template works
2. Some idea of library/language specific client code that you are generating as the generated code should be as natural (not spaghetti) as possible for someone who is already familiar with the language/library.
3. You need to also verify the generated code and write appropriate test cases to cover the various model input scenarios.

Related Link - https://github.com/foss42/apidash/discussions/80

animator:
@CodeHariK yes

animator:
Sure you can refer to the existing codegens and send across a PR for review when ready.

apoorvdwi:
@ashitaprasad @animator is this issue still open to work on?

PR #163, opened 2024-02-23
Feat : Add CodeGen for Go (http)

CodeHariK:
Feat : Add CodeGen for Go (http) #130

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ API Dash currently supports API integration code generation for the following la
 | cURL                   |               |
 | HAR                    |               |
 | Dart                   | `http`        |
+| Go                     | `net/http`    |
 | JavaScript             | `axios`       |
 | JavaScript             | `fetch`       |
 | JavaScript (`node.js`) | `axios`       |
--- a/lib/codegen/codegen.dart
+++ b/lib/codegen/codegen.dart
@@ -1,3 +1,4 @@
+import 'package:apidash/codegen/go/nethttp.dart';
 import 'package:apidash/models/models.dart' show RequestModel;
 import 'package:apidash/consts.dart';
 import 'dart/http.dart';
@@ -25,6 +26,8 @@ class Codegen {
         return DartHttpCodeGen().getCode(requestModel, defaultUriScheme);
       case CodegenLanguage.dartDio:
         return DartDioCodeGen().getCode(requestModel, defaultUriScheme);
+      case CodegenLanguage.goNetHttp:
+        return GoNetHttpCodeGen().getCode(requestModel, defaultUriScheme);
       case CodegenLanguage.jsAxios:
         return AxiosCodeGen().getCode(requestModel, defaultUriScheme);
       case CodegenLanguage.jsFetch:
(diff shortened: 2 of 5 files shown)


animator:
@CodeHariK Replace all tabs by whitespaces.

CodeHariK:
Where

CodeHariK:
I changed tabs to spaces in go code

CodeHariK:
```dart
import 'package:http/http.dart' as http;

void main() async {
  var uri = Uri.parse('https://api.foss42.com');

  final response = await http.get(uri);

  int statusCode = response.statusCode;
  if (statusCode >= 200 && statusCode < 300) {
    print('Status Code: $statusCode');
    print('Response Body: ${response.body}');
  } else {
    print('Error Status Code: $statusCode');
    print('Error Response Body: ${response.body}');
  }
}
```

```go
package main

import (
  "fmt"
  "io"
  "net/http"
)

func main() {
  url := "https://api.foss42.com"
  method := "GET"
  headers := map[string]string{}

  req, err := http.NewRequest(method, url, nil)
  if err != nil {
    fmt.Println("Error creating request:", err)
    return
  }

  // Set headers
  for key, value := range headers {
    req.Header.Set(key, value)
  }

  // Make the HTTP request
  client := &http.Client{}
  resp, err := client.Do(req)
  if err != nil {
    fmt.Println("Error making request:", err)
    return
  }
  defer resp.Body.Close()

  fmt.Printf("%s\n%s\n\n", resp.Status, resp.Header)

  // Read and print the response
  bodyText, err := io.ReadAll(resp.Body)
  if err != nil {
    fmt.Println("Error reading response body:", err)
    return
  }
  fmt.Println(string(bodyText))
}
```

These two are equivalent dart and go code

Go code is generally lengthy, unlike most languages, but it is totally valid, I don't see any problem or difference in any test case.

animator:
@CodeHariK the difference between both the codes is very evident. It is not about the length, but about what is required and what is not. Let me know when you have made the required changes to the PR.

CodeHariK:
Please verify

ashitaprasad:
@CodeHariK, Deleting comments is not how open source works. This is a completely unacceptable behavior. Others who might come to the PR will now have no clue about the flow of discussions that were happening. Open source is about being open to feedback, accepting mistakes and improving. Hope you respect the spirit of open source from next time.

CodeHariK:
bro it was useless comment, in which i was comparing my old code with dart code and another one in which i added spaces instead of tabs, and i am new to this, i didn't know.

ashitaprasad:
@CodeHariK The comments were important to the conversation at that time, post which the PR was modified. Each comment in this discussion is part of the PR history and just because any comment is no longer relevant does not mean you should delete it, as that comment led to the improvement.

CodeHariK:
What am i supposed to do now

ashitaprasad:
Just don't repeat it again in this project or any other open source project.

CodeHariK:
Ok, Please verify

animator:
@CodeHariK Can you kindly update the test cases to include example Request Models POST 4 to POST 9 which have been recently added.

ashitaprasad:
Thanks for working on this PR.
This PR currently has issues with multipart. Hence, we have accepted an alternate PR which resolves the issue.

PR #174, opened 2024-02-23
feat: add codegen for rust(actix)

### LINKED ISSUE #153: Add CodeGen for Rust (Actix Web)

Add CodeGen for Rust
Framework - https://github.com/actix/actix-web

Tanish2002:
Closes #153 

As discussed in the issue, actix client doesn't provide any inbuilt way make multiform requests like reqwest does.
Therefore I've made my own implementation of multiform requests similar to python(request) codegen.

Due to the nature of rust. 
If I have 2 url params, 2 headers, and 1 form data key value pair. Then the generated code is around 90-100 lines.

I could trim the lines down by using some third party crate(package) but I don't think we would want to include a lot of dependencies in the codegen.

Note: Currently these are the 3rd party packages that are used in codegen.
- awc (The actual actix client library)
- actix_rt (runtime for actix, they provide it in a separate crate)
- ~~serde (for deserializing structs)~~
- serde_json (for converting/serializing json to structs)
- mime_guess (for finding the mime type of file in multiform upload)

Edit: removed serde from deps


Tanish2002:
@ashitaprasad @animator If you approve of these changes then I can write tests for them.

Tanish2002:
Ok, so I've done a lot of jinja magic wherever possible to remove the extra code.
Now If I have 2 url params, 2 headers, and 1 form data key value pair. Then the generated code is around 50-70 lines.

animator:
@Tanish2002 Can you kindly update the test cases to include example Request Models POST 4 to POST 9 which have been recently added.

Tanish2002:
@animator done!

ashitaprasad:
LGTM 🚀

PR #200, opened 2024-02-25
textfield alignment

PCoder23:
There was an alignment issue in response screen search text field
before it looked like this 
Screenshot 2024-02-26 040239

now it looks like this 
Screenshot 2024-02-26 041106

is this fine ?


ashitaprasad:
Instead of directly raising a PR, you should first raise an issue in the provided format which includes the platform on which this issue has been observed.

PCoder23:
Ah, I will be carful about that :sweat_smile:

PR #209, opened 2024-02-26
Maximixe the window on double tap of title bar

### LINKED ISSUE #193: Double click to maximise/minimise on windows
Native windows apps usually have a "double click to maximise/unmaximise the window" feature.

Double click to maximise/unmaximise.

BrawlerXull:
## PR Description
Maximixe the window on double tap of title bar

## Related Issues

- Related Issue #
- Closes #193 

### Checklist

## Added/updated tests?

- [] Yes
- [x] I need help with writing tests

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -131,7 +131,7 @@ class DashApp extends ConsumerWidget {
           : Stack(
               children: [
                 kIsLinux ? const Dashboard() : const App(),
-                if (kIsWindows)
+                if (kIsWindows || kIsMacOS)
                   SizedBox(
                     height: 29,
                     child: WindowCaption(
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -18,17 +18,8 @@ void main() async {
     await setupWindow(sz: win.$1, off: win.$2);
   }
   runApp(
-    ProviderScope(
-      child: GestureDetector(
-        onDoubleTap: () async {
-          if (await windowManager.isMaximized()) {
-            windowManager.unmaximize();
-          } else {
-            windowManager.maximize();
-          }
-        },
-        child: const DashApp(),
-      ),
+    const ProviderScope(
+      child: DashApp(),
     ),
   );
 }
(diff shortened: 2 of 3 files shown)


MrunalrajRedij:
@BrawlerXull  `WindowCaption()` is inside `Stack()` and since you enabled it for MAC, now There will be an overlap of `WindowCaption()`  on MAC's native title bar. You can check that by wrapping `WindowCaption()` inside `Container()` and adding `color` to it. 
Now, this is not an issue for Windows because buttons are on the right side, but for MAC it's on the left side so It might cause unexpected issues.

Test: try double-clicking any of the three buttons as fast as possible. See if the button performs its own tasks or if it `unmaximizes `or `maximizes` the window.

MrunalrajRedij:
I don't have access to my MAC right now to test this.
The reason I brought this up is that, while implementing this in Windows I faced a similar issue and even you will face the same issue with this implementation.

As shown in the video below, if you double-click the `save` or `new `button it will `unmaximize `or `maximize`. 
Since WindowCaption() is on top of those buttons. (Red Color title bar shown in Image).

windowIssue

https://github.com/foss42/apidash/assets/75795149/eb854bb4-d026-42d0-a0bd-ab0524bc05a7

Was just curious if a similar thing was happening in MAC. On single click buttons are working correctly. The problem is when someone double clicks the button then there is an issue. 
This is not a major issue but still, can we find a workaround for this?

MrunalrajRedij:
What if we shift those two buttons (Save and New) a little down ? by giving a top padding of 29, which is the height of the title bar.

MrunalrajRedij:
Okay, got it so no issue in MAC and I can see why that is the case right, because the `Save `and `New` buttons are below the title bar and title bar is not covering the buttons.

And yes I implemented your code in Windows and the problem persists.

MrunalrajRedij:
> What if we shift those two buttons (Save and New) a little down ? by giving a top padding of 29, which is the height of the title bar.

How about this implementation?

animator:
Adding `onDoubleTap` to `GestureDetector` is not the correct approach so closing this PR.
Please open a new PR in case you come up with a better approach.

PratyushChauhan:
If we look at the widget tree, the sizedbox that has WindowCaption as parent occupies more space than necessary causing the undesired behavior while double clicking "Save" . 

A Solution would be to resize the width of the Sizedbox containing WindowCaption.

Is this the correct approach @animator ?

image

PR #210, opened 2024-02-26
Feature/add websocket support

### LINKED ISSUE #15: Feature request: Websocket Support
I would love to have Websocket support.
Is this on your roadmap?

sebastianbuechler:
## Add WebSocket support

image

This feature adds basic WebSocket support to apidash. It is according to the sketches in #15, but the headers are not yet incorporated into the web socket connection. Including the body does not make sense for WebSockets if I'm not mistaken, so we should remove it as well.

My proposal for the query params and the headers: If they get changed, the connection should be dropped and the user needs to manually reconnect, as the server connection will change. Maybe the UI should be adapted for this.

Thus open points:
 - [x] Remove body tab
 - [x] Include query params into WebSocketManager
 - [x] Handle query param changes into an existing connection
 - [ ] Include headers into WebSocketManager
 - [ ] Handle header changes into an existing connection

Testing is a bit harder for WebSockets to begin with. Maybe inject an HTTP client manually to mock it?

## Related Issues

- Closes #15

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] I need help with writing tests


sebastianbuechler:
@ashitaprasad In contrast to #215 this works with the same request model and dedicated WebSocket states. This allows for simultaneous connections to WebSocket servers. Let me know if adaptions are needed or further clarification on the specs.

animator:
@sebastianbuechler Kindly make this change https://github.com/foss42/apidash/issues/307

PR #277, opened 2024-03-11
added code generation language selector when error

### LINKED ISSUE #275: Missing code generation language selector
Whenever a code generation fails, the programming language selector gets hidden and raise issue appears. Every time I have to go to settings and change the language.

The programming language selector should not be hidden and their should be a retry button in place of copy and download icons so that the user can retry generating code from there.

_Your response here_

Tilakraj-B:
## PR Description
prior we were checking if the code generated in code_pane is null or not and returning the error widget without adding the code generator language changing bar . I removed the check in code_pane and added the check in the ViewCodePane(class) which render the error with the code generator language selector and replace the copy and download button with a single button Retry.

## Related Issues
- Related Issue #275 
- Closes #275 

### Checklist

## Added/updated tests?
- [x] Yes for the RetryButton Widget

### DIFF
--- a/lib/consts.dart
+++ b/lib/consts.dart
@@ -1,9 +1,10 @@
-import 'dart:io';
 import 'dart:convert';
+import 'dart:io';
+
+import 'package:davi/davi.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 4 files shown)


ashitaprasad:
Closing the PR as the issue has been resolved.
Check issue thread for more details.

Issue #309, opened 2024-03-16
Checkbox should not be selected if empty


animator:
**Describe the bug/problem**
The checkbox for URL Parameters & Headers should not be selected if no data has been added.

![Screenshot 2024-03-16 at 5 52 27 AM](https://github.com/foss42/apidash/assets/615622/8ebf73b1-3465-4107-892c-ea9760e627f5)

PR #300, opened 2024-03-14
fix: missing drag scrolling

DenserMeerkat:
## PR Description

Have used the [data_table_2](https://pub.dev/packages/data_table_2) and matched the current look

davi  | data_table_2 
--- | --- 
current_table | data_table_2

Fixed the text obscuring issue
current  | updated
--- | --- 
current_formData | new_form

## Related Issues

- Related Issue #299,
#250, 
#350 
#33 
- Closes #299,
#250 
#350 
### Checklist
- [X] I have gone through the [contributing guide](https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md)
- [X] I have run the tests (`flutter test`) and all tests are passing

## Added/updated tests?
_We encourage you to add relevant test cases._

- [ ] Yes
- [ ] No, and this is why: _please replace this line with details on why tests have not been included_

PR #313, opened 2024-03-16
feat: automatic row adding, handling checkbox selection when empty

### LINKED ISSUE #304: Relocate/Remove "+ Add Header", "+ Add Param" & "+ Add Form Data" Buttons
First discussed in #58 

"+ Add Header", "+ Add Param" & "+ Add Form Data" buttons require users to go down to the screen, which at times is inconvenient. These buttons can be relocated to a more accessible position like above the parameter fields.

It would also be a much better UX if a new empty row is automatically added below the last row if the user enters something in the last row. In this case the buttons can be eliminated.

DenserMeerkat:
## PR Description

>**_Chose to merge fixes for issues #304 and #309 into a single PR to avoid potential conflicts during merging as they required modifying the same file._**

Rows now get added automatically when user enters anything in the last row, accordingly removed the Add Param/Header/Form Data buttons.

Mimicked the Postman's behaviour of adding rows automatically and of checking and unchecking rows.
Postman hides the Checkbox if a field isn't touched yet, but I decided to keep it and have it disabled since hiding it didn't look good. Expecting feedback on this.

## Related Issues

- Related Issue #304 
#309 
- Closes #304 
#309
### Checklist

## Added/updated tests?

- [ ] Yes
- [X] No, and this is why: I Don't believe it requires/breaks any tests, will update if asked to.

### DIFF
--- a/lib/screens/home_page/editor_pane/details_card/request_pane/request_form_data.dart
+++ b/lib/screens/home_page/editor_pane/details_card/request_pane/request_form_data.dart
@@ -17,7 +17,7 @@ class FormDataWidget extends ConsumerStatefulWidget {
 class _FormDataBodyState extends ConsumerState<FormDataWidget> {
   late int seed;
   final random = Random.secure();
-  late List<FormDataModel> rows;
+  late List<FormDataModel> formRows;
 
   @override
   void initState() {
@@ -28,7 +28,7 @@ class _FormDataBodyState extends ConsumerState<FormDataWidget> {
   void _onFieldChange(String selectedId) {
     ref.read(collectionStateNotifierProvider.notifier).update(
           selectedId,
-          requestFormDataList: rows,
+          requestFormDataList: formRows.sublist(0, formRows.length - 1),
         );
   }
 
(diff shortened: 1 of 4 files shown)


animator:
@DenserMeerkat Check out the case below for which it is breaking. Add params/headers -> save -> close -> reopen -> no empty row at the bottom.

https://github.com/foss42/apidash/assets/615622/e18f3a48-8043-45f9-8545-7a35d535735a

Also, after further pondering we think it is better to retain the buttons in the bottom for a while until this approach works perfectly in the wild.

DenserMeerkat:
>Check out the case below for which it is breaking. Add params/headers -> save -> close -> reopen -> no empty row at the bottom.

Will look into it.

>Also, after further pondering we think it is better to retain the buttons in the bottom for a while until this approach works perfectly in the wild.

Right, understandable.

animator:
@DenserMeerkat Just a clarification in case the previous comment caused any confusion ... 
We want to include the features in this PR + retain the buttons as well.

DenserMeerkat:
> @DenserMeerkat Just a clarification in case the previous comment caused any confusion ... We want to include the features in this PR + retain the buttons as well.

Oh gotcha, will add them back.

ashitaprasad:
@DenserMeerkat Is the PR ready for review ?

DenserMeerkat:
Not yet @ashitaprasad,
I'm still working on the breaking case @animator pointed out.

[#300](https://github.com/foss42/apidash/pull/300) is ready though.

ashitaprasad:
Sure 👍

DenserMeerkat:
The PR is now ready for review @ashitaprasad.
The empty row now shows up when there are already existing rows in a request.

animator:
LGTM 🚀

PR #318, opened 2024-03-16
Add Curl Rust Code Generator

### LINKED ISSUE #151: Add CodeGen for Rust (curl-rust)

Add CodeGen for Rust 
Library - https://github.com/alexcrichton/curl-rust

apoorvdwi:
## PR Description

Added curl rust code generator and corresponding test file

## Related Issues

- Related Issue #151 
- Closes #151 

### Checklist

## Added/updated tests?
- [x] Yes


apoorvdwi:
@ashitaprasad @animator Please review this and let me know if any changes need to be done

Some pointers:
- rust curl does not have a native query method that can be used to add query params so added it in the URL only
- rust curl also does not provide method-specific functions for DELETE, PATCH and HEAD requests. The way I've implemented it is the way to make those calls.

I have tested the code generated for test cases on code sandbox and it is working as expected

lib/codegen/rust/curl_rust.dart
```diff
@@ -0,0 +1,144 @@
...
+""";
+
+  String kTemplateFormData = """
+  let mut form = curl::easy::Form::new();
+  {% for field in fields %}
+  form.part("{{field.name}}")
+    {% if field.type == "file" %}.file("{{field.value}}"){% else %}.contents(b"{{field.value}}"){% endif %}
+    .add();
+  {% endfor %}
+  easy.httppost(form);
```
Tanish2002, line 53: this should be easy.httppost(form).unwrap() since httppose returns a `Result<(), Error>` and if you don't add a unwrap() or check for errors manually the compiler will throw a warning.

lib/codegen/rust/curl_rust.dart
```diff
@@ -0,0 +1,144 @@
...
+  """;
+
+  String kTemplateHeader = """
+  {% if headers %}let mut list = List::new();{% for header, value in headers %}
+  list.append("{{header}}: {{value}}");{% endfor %}
```
Tanish2002, line 58: this should be `list.append("{{header}}: {{value}}").unwrap()` since append returns a `Result<(), Error>` and if you don't add a unwrap() or check for errors manually the compiler will throw a warning.

Tanish2002:
also `use curl::easy::List;` is imported even when there are no headers in the codegen.
for example check your get 12 test case

apoorvdwi:
Hey @Tanish2002 
Thanks for pointing out the changes. I have fixed the mentioned issues 😄

Tanish2002:
LGTM! 
The maintainers will give their review soon😄

ashitaprasad:
In some of the post cases like post4, post7 etc. getting unused import warning. Please fix it

```
warning: unused import: `curl::easy::List`
 --> src/post7.rs:2:5
  |
2 | use curl::easy::List;
  |     ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

apoorvdwi:
@ashitaprasad Fixed it. This was happening because formdata content type header was not being added.

ashitaprasad:
@apoorvdwi Do we need to send the formdata content type header here in this codegen? Is it not automatically added by the library? If you are providing it, why are you not adding any boundary?

apoorvdwi:
curl rust does not explicitly sets the header, we have to do it. But since we are putting data in a easy form which is native supported we do not need a boundary here.

animator:
@apoorvdwi For json and text we need to set the relevant content-type header, but `multipart-formdata` always has additional boundary information. If it is natively supported then it means that the header is also being created automatically and what you are providing explicitly is not being used.

Print the request headers that are being sent with and without your explicit setting of `multipart-formdata` header and paste it in the comment below.

apoorvdwi:
Got it. I checked the request with explicit content type and without it. It is coming in both the cases. My bad I got a bit confused.

I have fixed the headers to remove formdata type and updated test cases as well

animator:
@apoorvdwi I clearly said above

> Print the request headers that are being sent with and without your explicit setting of multipart-formdata header and paste it in the comment below.

Share the outputs for both the cases.

apoorvdwi:
Without content type set
```
{
  "content-length": "330",
  "accept": "*/*",
  "content-type": "multipart/form-data; boundary=------------------------273a16fce4d4da7d",
  "x-forwarded-for": "178.63.141.96",
  "x-forwarded-host": "apoorv.free.beeceptor.com",
  "x-forwarded-proto": "https",
  "accept-encoding": "gzip"
}
```

With content type
```
{
  "content-length": "330",
  "accept": "*/*",
  "content-type": "multipart/form-data; boundary=------------------------601231ad2c4a841c",
  "x-forwarded-for": "178.63.141.96",
  "x-forwarded-host": "apoorv.free.beeceptor.com",
  "x-forwarded-proto": "https",
  "accept-encoding": "gzip"
}
```

animator:
@apoorvdwi 

Why are you setting these headers?
"accept": "*/*",
"x-forwarded-for": "178.63.141.96",
"x-forwarded-host": "apoorv.free.beeceptor.com",
"x-forwarded-proto": "https",
"accept-encoding": "gzip"

Do the following exercise for [request model POST 4](https://github.com/foss42/apidash/blob/6bbbe9cc1e5b236886c466b186040186341fd8d9/test/request_models.dart#L236C7-L236C24):
- Generated codes (with and without your explicit setting of multipart-formdata header)
- In the code add lines to print the request headers that are being sent.

Share the two codes below. I will run it on my PC and verify.

apoorvdwi:
I'm not setting those headers
I'm running it from code sandbox and they are going through proxies to `apoorv.free.beeceptor.com` which is why headers are set.

Here are the codes - 
Without content type
```rust
use curl::easy::Easy;

fn main() {
    let mut easy = Easy::new();
    let mut data = Vec::new();
    easy.url("https://apoorv.free.beeceptor.com").unwrap();
    easy.post(true).unwrap();

    let mut form = curl::easy::Form::new();

    form.part("text").contents(b"API").add().unwrap();

    form.part("sep").contents(b"|").add().unwrap();

    form.part("times").contents(b"3").add().unwrap();

    easy.httppost(form).unwrap();

    {
        let mut transfer = easy.transfer();
        transfer
            .write_function(|new_data| {
                data.extend_from_slice(new_data);
                Ok(new_data.len())
            })
            .unwrap();
        transfer.perform().unwrap();
    }

    let response_body = String::from_utf8_lossy(&data);

    println!("Response body:{}", response_body);
    println!("Response code: {}", easy.response_code().unwrap());
}
```

With content type
```rust
use curl::easy::Easy;
use curl::easy::List;

fn main() {
    let mut easy = Easy::new();
    let mut data = Vec::new();
    easy.url("https://apoorv.free.beeceptor.com").unwrap();
    easy.post(true).unwrap();

    let mut form = curl::easy::Form::new();

    form.part("text").contents(b"API").add().unwrap();

    form.part("sep").contents(b"|").add().unwrap();

    form.part("times").contents(b"3").add().unwrap();

    let mut list = List::new();
    list.append("Content-Type: multipart/form-data").unwrap();
    easy.http_headers(list).unwrap();

    easy.httppost(form).unwrap();

    {
        let mut transfer = easy.transfer();
        transfer
            .write_function(|new_data| {
                data.extend_from_slice(new_data);
                Ok(new_data.len())
            })
            .unwrap();
        transfer.perform().unwrap();
    }

    let response_body = String::from_utf8_lossy(&data);

    println!("Response body:{}", response_body);
    println!("Response code: {}", easy.response_code().unwrap());
}
```

animator:
> In the code add lines to print the request headers that are being sent.

@apoorvdwi You have not added the lines like I said above.

apoorvdwi:
I'm not sure how to print request headers from the code, don't see any native function in curl-rust, on the receiving side it can be done which is why I used a mock server

animator:
Then why did you share [this](https://github.com/foss42/apidash/pull/318#issuecomment-2010654394) in first place when I repeatedly asked you to print the headers that are being sent by the code.
This is simply not an ethical behaviour where instead of being honest in the first place you are simply wasting the time of maintainers who put in so much effort to provide the right feedback, just to get your PR accepted.
Take this as an assignment.
I am closing this PR. Once you have completed the assignment you can share it in this thread and raise a new PR.

apoorvdwi:
@animator I think you misunderstood what I said
I used a mock server located at `apoorv.free.beeceptor.com`
I ran the code in the code sandbox and sent the request there.
The headers I pasted are what the server received
Here are the screenshots
Screenshot 2024-03-21 at 3 27 41 AM
Screenshot 2024-03-21 at 3 28 19 AM

Issue #22, opened 2023-05-05
Enhance Request Body Editor: JSON formatting, syntax highlighting, validation 


animator:
This feature was highlighted in #9 

Currently, the Request Body is a plain text editor. In case the body is of type `json`, the editor can provide additional features (without hampering performance) like:
- JSON formatting
- Syntax highlighting
- JSON validation

malik-vishu:
@animator Could you explain it a bit more? With screenshots if poosible.

ashitaprasad:
Hi @malik-vishu,
Please find attached screenshots 

## JSON formatting 

<img width="1218" alt="json formatting" src="https://github.com/foss42/apidash/assets/1382619/3553bed2-631f-4a69-9f2f-63f6729513fc">

## Syntax Highlighting

<img width="1038" alt="json syntax highlighting" src="https://github.com/foss42/apidash/assets/1382619/127d7c96-7643-4529-abd0-0fe39f5f64e4">

## JSON validation

<img width="996" alt="json validation" src="https://github.com/foss42/apidash/assets/1382619/b3c31783-6d8b-4fda-b5cd-ce87dd9ce4a9">

antonioCaballero17:
Hi @animator, @ashitaprasad , I am currently working in this feature,

ashitaprasad:
Sure @antonioCaballero17 waiting for your PR.
If you have any doubt/query feel free to post it in our discord channel [#foss-apidash](https://bit.ly/fgfossd)

antonioCaballero17:
Based on this issue i just develop and published this pacakge
[https://github.com/antonioCaballero17/json_text_field](https://github.com/antonioCaballero17/json_text_field
)Can I use this in apiDash?

animator:
@antonioCaballero17 you can send across a PR. Will test it out.

antonioCaballero17:
Please check #103

antonioCaballero17:
Hi! Sorry for my ghosting, i was moving home last month... I see that you merge the PR. So, great! feature 22 is done!
I am looking forward for other feature, i am going to check the issues

animator:
@antonioCaballero17 the PR was merged because it offers something very close to the solution that someone can pick up to continue working on. See the TODO [here](https://github.com/foss42/apidash/blob/d346be2a54f42e21ab572d4d26a3086bfdb043c1/lib/screens/home_page/editor_pane/details_card/request_pane/request_body.dart#L42)
This issue is still open and requires work.

[...]

PR #358, opened 2024-03-25
Feat syntax highlight request body

Tanish2002:
## PR Description

This PR Introduces syntax highlighting for JSON type request body. 
We had a comment referencing the incomplete [`JsonTextFieldEditor` implementation](https://github.com/foss42/apidash/blob/f3a7865459e3f0b2b44b9012026fe71159d8897d/lib/screens/home_page/editor_pane/details_card/request_pane/request_body.dart#L42). Instead of hacking together our own, I've used a third-party package [code_field](https://github.com/BertrandBev/code_field). 

This package provides syntax highlighting as well as inserts tab characters at appropriate positions. 
We can also use this same package in the future for other request bodies such as Graphql, protobuf, etc. Once we support them.

I've also changed the default `TextFieldEditor` widget using the `plaintext` theme, as the name suggests, it doesn't provide any syntax highlighting but does provide simple tab character insertion wherever needed.

Here is a list of currently supported languages and themes that are supported: [languages](https://github.com/git-touch/highlight.dart/tree/master/highlight/lib/languages), [themes](https://github.com/git-touch/highlight.dart/tree/master/flutter_highlight/lib/themes)

Lastly Here are some screenshots:
image
image

I've decided to use `monokaiSublimeTheme` for dark mode and `defaultTheme` for light mode since they look the most natural for our colorscheme.

If unsatisfied with the current choice please tell me any other theme from the list above, The developer has also created a nice site to figure out a theme: [Here](https://bertrandbev.github.io/code_field/)

### Checklist

## Added/updated tests?
- [x] Yes

### DIFF
--- a/lib/screens/home_page/editor_pane/details_card/request_pane/request_body.dart
+++ b/lib/screens/home_page/editor_pane/details_card/request_pane/request_body.dart
@@ -39,10 +39,9 @@ class EditRequestBody extends ConsumerWidget {
             child: switch (contentType) {
               ContentType.formdata =>
                 const Padding(padding: kPh4, child: FormDataWidget()),
-              // TODO: Fix JsonTextFieldEditor & plug it here
               ContentType.json => Padding(
                   padding: kPt5o10,
-                  child: TextFieldEditor(
+                  child: JsonTextFieldEditor(
                     key: Key("$selectedId-json-body"),
                     fieldKey: "$selectedId-json-body-editor",
                     initialValue: requestModel?.requestBody,
--- a/lib/widgets/editor.dart
+++ b/lib/widgets/editor.dart
@@ -1,7 +1,9 @@
-import 'dart:math' as math;
+import 'package:code_text_field/code_text_field.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:apidash/consts.dart';
+import 'package:flutter_highlight/themes/default.dart';
+import 'package:flutter_highlight/themes/monokai-sublime.dart';
+import 'package:highlight/languages/plaintext.dart';
 
 class TextFieldEditor extends StatefulWidget {
   const TextFieldEditor({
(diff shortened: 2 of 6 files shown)


animator:
We already have dark & light syntax highlighting themes for API Dash which is used in codegens here -https://github.com/foss42/apidash/blob/f3a7865459e3f0b2b44b9012026fe71159d8897d/lib/consts.dart#L160

animator:
You need to go through the relevant issue #22 which talks about the deliverables and the past work & the issues in the WIP PR that was merged.

Tanish2002:
@animator As per the issue, I believe syntax highlight is taken care of through this PR.

I should be able to add formatting relatively easily(using `kEncoder.convert`), and validation should also be possible.
However, I would need some help with the UI.

I added a "format JSON" elevated button. It looks like this:
image
Do you think this is fine? or I should do something else. 


Also, where do I keep the validation error? I'm thinking of simply adding a check to the OnChanged hook that will check if JSON is valid or not. If not then it would show an error. But not sure where to place this widget in the UI

Tanish2002:
@animator This is what I've come up with. Haven't added any new packages, simply using the json function from `dart:convert` for formatting as well as validation issues.

https://github.com/foss42/apidash/assets/55488165/ce121489-1035-497b-9a8f-859d6dff44cb

Tanish2002:
@animator pushed the new changes, made it look pretty as much as I could 😅.
You can check the following video for a example.
https://github.com/foss42/apidash/assets/55488165/22ebac0d-eb0c-4ab6-bf4f-b674f1fc43ab

animator:
The original text editor is perfectly fine. Why was it modified?
Only JSON editor requires modification.

Tanish2002:
> The original text editor is perfectly fine. Why was it modified? Only JSON editor requires modification.

Yeah,
 
- I modified the editor class to ensure consistency between all editor widgets and make it easier for other developers to understand.
- I used plaintext as the language to remove highlighting.
https://github.com/foss42/apidash/blob/4912f71a77159a67704e413319b6ef5b90ca5e7a/lib/widgets/editor.dart#L33
- I believe in using existing packages (which I added in this PR) for consistent code rather than creating custom implementations for every editor type.
- I haven't implemented this but I think an abstract Editor class (using the `CodeField` widget) offers flexibility. Subclasses for various editors(json, graphql, protobuf, etc) can then add features like formatting and validation as needed.

@animator If you believe this isn't what we should do, then I can go ahead and revert the changes for that.

animator:
Closed by https://github.com/foss42/apidash/pull/595

Issue #118, opened 2024-02-22
Add UI/riverpod tests cases


ashitaprasad:
**Tell us about the task you want to perform and are unable to do so because the feature is not available**
Currently, API Dash has Unit tests & Widget tests, but needs support for adding UI tests which involves injecting riverpod states and testing the behaviour of screens. This will help improve the code coverage.

sixtusagbo:
I'll like to work on this.

I went through the contribution guidelines and the second step is to comment on the issue. I want to fork the repo and run it locally now.

sixtusagbo:
I've run the app and the tests and I'm getting started on the UI tests now

ashitaprasad:
Awesome @sixtusagbo, you can send across a PR once you are done writing tests.

animator:
Awesome @sixtusagbo 👍

sixtusagbo:
I've written tests for some of the providers. We'll be on break during the GSoC period so I have all the time to work on this project.

ashitaprasad:
@sixtusagbo Please send across a PR as soon as possible so that we can review the approach and provide feedback.

sixtusagbo:
I've done that, while I wait. I want to round up the Enter Key creates a new entry for "Add Param" and "Add Header"

animator:
@sixtusagbo that issue has already been resolved and is in the final stages of review. Kindly work on adding more UI tests.

sixtusagbo:
Okay

Tanish2002:
@ashitaprasad @animator This issue can be closed now I believe.

sixtusagbo:
I don't think so. There are other test cases I'm working on. I just haven't updated remote yet. And I have an approach for the mock. I had to modify the `request()` helper method in `http_services.dart`

Tanish2002:
Ah my bad, the linked PR was closed so I thought it was fine.

sixtusagbo:
I'll open a PR on my progress before the end of the week. I added more tests. My goal is to finish saveDataStateProvider, clearDataStateProvider, hasUnsavedChangesProvider, nameTextFieldFocusNodeProvider and searchQueryProvider before the actual GSoC coding week begins.

PR #361, opened 2024-03-26
Improve http service

sixtusagbo:
## PR Description

I was trying to mock the request made on the app by mocking the `http.request()` method because parts of it that makes the request using `http.Client` object and it needs to be replaced to avoid making actual requests in the case of unit and widget tests.

I went through the code in `http_service.dart` and it does not accept as a parameter `http.Client` object. And `collectionStateNotifierProvider` does not accept `http.Client` object as a parameter too.
If it accepts a client object (as a parameter), anyone will be able to override the client when testing in order mock the http request or to unit test the service `http_service.dart`.

## Related Issues

- Related Issue #118 

### Checklist

## Added/updated tests?

- [x] Yes

### DIFF
--- a/test/providers/ui_providers_test.dart
+++ b/test/providers/ui_providers_test.dart
@@ -3,14 +3,23 @@ import 'dart:io';
 import 'package:apidash/providers/providers.dart';
 import 'package:apidash/screens/dashboard.dart';
 import 'package:apidash/screens/home_page/collection_pane.dart';
+import 'package:apidash/screens/home_page/editor_pane/details_card/code_pane.dart';
+import 'package:apidash/screens/home_page/editor_pane/details_card/response_pane.dart';
+import 'package:apidash/screens/home_page/editor_pane/editor_default.dart';
+import 'package:apidash/screens/home_page/editor_pane/editor_pane.dart';
+import 'package:apidash/screens/home_page/editor_pane/url_card.dart';
 import 'package:apidash/screens/home_page/home_page.dart';
 import 'package:apidash/screens/intro_page.dart';
 import 'package:apidash/screens/settings_page.dart';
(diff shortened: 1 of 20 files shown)


ashitaprasad:
We will currently not make any changes to the http_service. Instead of mocking the request you can send an actual request in the tests as these apis are also maintained by us.

sixtusagbo:
Okay, I've removed the mocks. I am sending actual requests now. I've completed test cases for `codePaneVisibleStateProvider`

test/providers/ui_providers_test.dart
```diff
@@ -336,4 +343,274 @@ void main() {
...
+
+      // Verify that NotSentWidget is visible
+      expect(find.byType(NotSentWidget), findsOneWidget);
+
+      // Add some data in URLTextField
+      Finder field = find.descendant(
+        of: find.byType(URLField),
+        matching: find.byType(TextFormField),
+      );
+      const url = 'https://api.foss42.com/country/codes';
```
animator, line 544: Put URL const in https://github.com/foss42/apidash/blob/main/test/test_consts.dart

animator:
Just like we have defined example request models in https://github.com/foss42/apidash/blob/main/test/request_models.dart, you need to add constants like the URL used in tests in https://github.com/foss42/apidash/blob/main/test/test_consts.dart
Also, use `api.apidash.dev` not `api.foss42.com`

test/providers/ui_providers_test.dart
sixtusagbo: I've added it

sixtusagbo: I've added it

sixtusagbo:
Okay Sir, I just updated the files now.

animator:
LGTM 🚀

PR #363, opened 2024-03-27
Fix hive persistence issue

### LINKED ISSUE #359: Saving state never terminates if persistent files cannot be accessed / modified
**Describe the bug/problem**
To store data API Dash requires 4 files, namely - 
1. apidash-data.hive
2. apidash-data.lock
3. apidash-settings.hive
4. apidash-settings.lock

Now in Windows, these files are generated inside `%USERPROFILE%/Documents` directory and in Linux, these are generated under `~/` directory. If these locations are protected, i.e. external programs cannot create files in these locations or modify existing files, then if API Dash tries to access these files, it throws an error and the UI never updates. 

Reproducing this issue can be achieved via many ways, i.e. revoking write access to the generated files and trying to save a request. A simpler way to reproduce this issue on Windows is to enable Ransomware Protection which will disallow external programs trying to generate files under some predefined directories (`%USERPROFILE%/Documents` being one of them) and then trying to save a request.

After clicking `save`, the saving state never terminates.

Screenshot 2024-03-25 154145

API Dash throws error in the terminal stating it cannot access the files user to store persistent data.

Screenshot 2024-03-25 153740

**Expected behavior**
User should be presented with an error message stating that the respective files could not be read / modified and file permissions needs to be updated by the user. 

An enhancement to the solution is to add a default location for persistent data and then give the user ability to load data and settings from custom location using file path. This will allow users to version control request templates and share the existing requests among developers.

 - OS: Windows 11

yashas-hm:
## PR Description
Added a directory path for hive to store files. The path is application support directory which will be accessible by the app irrespective of permissions by the OS firewall.

By default hive stores its files at `getApplicationDocumentsDirectory`

Use `getApplicationSupportDirectory` instead of `getApplicationDocumentsDirectory` to ensure read and write permissions.
https://github.com/hivedb/hive/blob/3b12c31a221f97f5ec86fe20df63515aeedf88f0/hive_flutter/lib/src/hive_extensions.dart#L16

Related issue
https://github.com/isar/hive/issues/1044

## Related Issues

- Related Issue #359 
- Closes #359 

### Checklist

## Added/updated tests?
- [x] No, and this is why: Not Required

### DIFF
--- a/lib/services/hive_services.dart
+++ b/lib/services/hive_services.dart
@@ -8,7 +8,7 @@ const String kKeyDataBoxIds = "ids";
 const String kSettingsBox = "apidash-settings";
 
 Future<void> openBoxes() async {
-  final supportDir = await getApplicationSupportDirectory();
+  final supportDir = await getLibraryDirectory();
   await Hive.initFlutter(supportDir.path);
   await Hive.openBox(kDataBox);
   await Hive.openBox(kSettingsBox);
--- a/lib/services/hive_services.dart
+++ b/lib/services/hive_services.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:hive_flutter/hive_flutter.dart';
+import 'package:path_provider/path_provider.dart';
 
 const String kDataBox = "apidash-data";
 const String kKeyDataBoxIds = "ids";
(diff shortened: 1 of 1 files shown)


animator:
In the documentation, it is written -
> Your app should not use this directory for user data files.
Apart from some configs, we are majorly storing user data so documents folder is the right location.

yashas-hm:
My approach to solve the solution was this as support folder is always accessible. The other option is to store in app cache directory else we can prompt the use that access to the particular folder is not possible please give API Dash permission to read and write to the folder.

The document folder would not be accessible only if the user has denied permissions or added an additional firewall which would not the case for most of the users.

yashas-hm:
> Apart from some configs, we are majorly storing user data so documents folder is the right location.

I think the documentation mentions the user data such as user details like name, phone etc. user sensitive details.

I guess using getLibraryDirectory would be more logical. 

Documentation Reference:
> Path to the directory where application can store files that are persistent,
> backed up, and not visible to the user, such as sqlite.db.

animator:
Nope. Closing this PR as this is not a path issue, but permission issue.

PR #380, opened 2024-04-06
Fuzzy header search

### LINKED ISSUE #207: Implement fuzzy string matching to get header suggestions

Currently, header suggestions are a sequence of strings which contain the given substring with no logic implemented for ordering. Algorithms like [fuzzy string search](https://redis.com/blog/what-is-fuzzy-matching/) etc. can be used to provide the sequence of matching strings with the most relevant item at the top.

PratyushChauhan:
## PR Description

Previously we were only checking if a pattern is  present in a header string. As a result you would not get suggestions if you made a typo.
BEFORE:
320200891-35e13df0-5a8a-49bd-adf9-a96b5e22d569

This PR attempts to fix this issue using [Fuzzy](https://pub.dev/packages/fuzzy) package.

AFTER:
320200720-079a90ca-80e6-41fe-ad90-6e37702410bb

## Related Issues

- Related Issue #207 
- Closes #207 

### Checklist

## Added/updated tests?
_All previous tests except one have been added for the new function._

- [x] Yes

### DIFF
--- a/lib/utils/header_utils.dart
+++ b/lib/utils/header_utils.dart
@@ -121,7 +121,10 @@ List<String> getFuzzyHeaderSuggestions(String pattern) {
   final keys = headers.keys.toList();
   final fuse = Fuzzy(keys,
       options: FuzzyOptions(
-        distance: 50,
+        distance: 40,
+        matchAllTokens: true,
+        isCaseSensitive: true,
+        tokenize: true,
         threshold: 0.4,
       ));
   final results = fuse.search(pattern);
--- a/lib/utils/header_utils.dart
+++ b/lib/utils/header_utils.dart
@@ -1,3 +1,4 @@
+import 'package:fuzzy/data/result.dart';
 import 'package:fuzzy/fuzzy.dart';
 
 Map<String, String> headers = {
@@ -121,13 +122,16 @@ List<String> getFuzzyHeaderSuggestions(String pattern) {
   final keys = headers.keys.toList();
   final fuse = Fuzzy(keys,
       options: FuzzyOptions(
-        distance: 40,
-        matchAllTokens: true,
-        isCaseSensitive: true,
+        distance: 10,
         tokenize: true,
-        threshold: 0.4,
+        matchAllTokens: true,
+        tokenSeparator: "-",
+        threshold: 0.2,
+        shouldSort: true,
+        minMatchCharLength: 1,
       ));
   final results = fuse.search(pattern);
-  final suggestions = results.map((result) => result.item as String).toList();
+  final List<String> suggestions =
+      results.map((result) => result.item as String).toList();
   return suggestions;
 }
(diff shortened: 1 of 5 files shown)


animator:
It should show recommendation if there is a typo but if an exact substring match exists, it should show only them.

PratyushChauhan:
> It should show recommendation if there is a typo but if an exact substring match exists, it should show only them.

@animator I have a suggestion to implement this.

in `getFuzzyHeaderSuggestions()`:
```
List<String> getFuzzyHeaderSuggestions(String pattern) {
  final keys = headers.keys.toList();
  final fuse = Fuzzy(keys,
      options: FuzzyOptions(
        distance: 10,
        tokenize: true,
        matchAllTokens: true,
        tokenSeparator: "-",
        threshold: 0.2,
        shouldSort: true,
        minMatchCharLength: 1,
      ));
  final results = fuse.search(pattern);
  final List<String> suggestions =
      results.map((result) => result.item as String).toList();
  return suggestions;
}
```
 when we obtain `suggestions` we can then iteratively sort them into 2 lists, "predictions" and "recommendations" {in case there is a typo}. 

Example: every suggestion that exactly `contains(pattern)` will be a "prediction", else it will go to "recommendations".
Then if `suggestions.length == 0` in which case the user made a typo, we will display the fuzzy "recommendations" list otherwise we will only display "predictions" that will contain the pattern substring.


Let me know if this is desirable.

PratyushChauhan:
Doing this also passes the one test case I had to edit. This should do better than the previous approach.

PratyushChauhan:
Passes all test cases and does not show approximate suggestions if no typo exists. If typo exists then it shows fuzzy suggestions.

ashitaprasad:
@PratyushChauhan For the case, when the user just types `a` the result should include headers beginning with `a` followed by other headers. This is not happening currently. 

Screenshot 2024-04-11 at 8 05 32 AM

PratyushChauhan:
@ashitaprasad worked on your suggestion, now the first character match is given preference in predictions.

image

animator:
@PratyushChauhan Why is `Accept` lower than `Accept-Charset` in your implementation?

PratyushChauhan:
@animator  this most likely happens when we re-order the predictions {in case of typo} to include first character matching strings near the top. 

```
// sort the predictions based on first character match
  for (int i = 0; i < predictions.length; i++) {
    if (predictions[i][0].toLowerCase() == pattern[0].toLowerCase()) {
      final String temp = predictions.removeAt(i);
      predictions.insert(0, temp); //push to front
    }
  }
```
In case of a typo each string that has a matching first character is pushed to position 0. Hence the order given by the fuzzy finder is reversed which makes `Accept` have a lower ranking than `Accept-Charset`.

A way to fix this would be to introduce a `currentIndex` variable to keep track of which index we need to push to. This will preserve the order given by the fuzzy finder package.

Fix: 
```
  int currentIndex = 0;
  // sort the predictions based on first character match
  for (int i = 0; i < predictions.length; i++) {
    if (predictions[i][0].toLowerCase() == pattern[0].toLowerCase()) {
      final String temp = predictions.removeAt(i);
      predictions.insert(currentIndex, temp); //push to front
      currentIndex++;
    }
  }
```

image

PratyushChauhan:
@animator @ashitaprasad  is this implementation fine?

PR #473, opened 2024-10-03
feat(cookie): Added the support for http cookie

### LINKED ISSUE #465: Add support for HTTP Cookies
I want to make a request to an endpoint that has a cookie session e.g.
```
curl --cookie "session_token=abcd" www.example.com/secured
```

Add an option to allow for cookies / persistent http client

-

userAdityaa:
## PR Description
This PR introduces support for sending HTTP cookies with API requests. The new functionality allows users to include cookies in their headers, which is especially useful when interacting with endpoints that rely on session tokens or other cookie-based authentication.

## Related Issues
- Closes #465 

### Checklist

## Added/updated tests?

- [ ] Yes


ashitaprasad:
The PR is incorrect as you have added Cookie support to a java Codegen, but we were looking for cookie support in the app. Also, cookies have to be retrieved from the reponse model headers of a request.

Closing it for now as this feature will require more discussion on the issue thread.

PR #488, opened 2024-11-01
feat(uriScheme): Added the support for url without http/https prefix

### LINKED ISSUE #487: Issue: Lack of Support for URL Scheme Without HTTP/HTTPS Prefix
### Tell us about the task you want to perform and are unable to do so because the feature is not available

I want to test APIs on my local server using the URL format localhost:3000 (without http://). Currently, API Dash does not support this format and returns an error, making it inconvenient to test APIs locally without explicitly adding the http:// prefix.

### Describe the solution/feature you'd like us to add

Add support for URLs using the localhost:3000 format, allowing users to enter URLs without needing the http:// scheme prefix. This feature would streamline the testing process by accepting both localhost:3000 and http://localhost:3000 as valid inputs, similar to how tools like Postman handle local URLs.

### Any other feedback you would like to provide regarding the site

API Dash is a great tool for testing APIs, and adding this small but useful feature would further enhance usability, especially for developers working in local development environments. Thank you for considering this improvement!

### Screenshot of the issue
Screenshot 2024-11-01 at 5 16 15 AM
Screenshot 2024-11-01 at 5 16 26 AM

userAdityaa:
## PR Description
This PR addresses the issue where URLs without an explicit HTTP or HTTPS scheme (such as localhost:3000) were not being recognized as valid requests. Previously, users needed to specify a full URL with a prefix (http:// or https://) to ensure the application processed it correctly. This PR implements a fallback mechanism to add a default scheme (HTTP) when a URL scheme is missing, improving usability and supporting more flexible URL inputs.
_Add your description_

## Related Issues
- Closes #487 

## Screenshot
Screenshot 2024-11-01 at 3 56 52 PM

### Checklist

### DIFF
--- a/packages/apidash_core/lib/consts.dart
+++ b/packages/apidash_core/lib/consts.dart
@@ -4,8 +4,13 @@ enum HTTPVerb { get, head, post, put, patch, delete }
 
 enum FormDataType { text, file }
 
-const kSupportedUriSchemes = ["https", "http"];
+enum SupportedUriSchemes { https, http }
+
+final kSupportedUriSchemes =
+    SupportedUriSchemes.values.map((i) => i.name).toList();
 const kDefaultUriScheme = "https";
+final kLocalhostRegex = RegExp(r'^localhost(:\d+)?(/.*)?$');
+
 const kMethodsWithBody = [
   HTTPVerb.post,
   HTTPVerb.put,
--- a/packages/apidash_core/lib/utils/uri_utils.dart
+++ b/packages/apidash_core/lib/utils/uri_utils.dart
@@ -29,9 +29,9 @@ String stripUrlParams(String url) {
   if (url == null || url == "") {
     return (null, "URL is missing!");
   }
-  final localhostRegex = RegExp(r'^localhost(:\d+)?(/.*)?$');
-  if (localhostRegex.hasMatch(url)) {
-    url = 'http://$url';
+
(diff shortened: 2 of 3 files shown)


userAdityaa:
Why is the defaultUriScheme set to https but not http ?

ashitaprasad:
@userAdityaa 
defaultUriScheme defaults to `https` as it is more oriented to test real endpoints.
Also, it is a user customizable property which can be changed via settings.

userAdityaa:
> @userAdityaa defaultUriScheme defaults to `https` as it is more oriented to test real endpoints. Also, it is a user customizable property which can be changed via settings.

Yes, I just show it after the my comment.

ashitaprasad:
The current approach is not correct and it will be better to have a regex based approach to check for localhost urls and take appropriate action.

userAdityaa:
> The current approach is not correct and it will be better to have a regex based approach to check for localhost urls and take appropriate action.

Ok, will look into it.

userAdityaa:
@ashitaprasad, Ready for review!

ashitaprasad:
testing is missing for this modification.

userAdityaa:
Hey @ashitaprasad, The test are added now. 
Screenshot 2024-11-02 at 6 56 58 PM

ashitaprasad:
Thanks for the PR 🚀

PR #497, opened 2024-11-11
feat:Cookies_support

Clasherzz:
## PR Description
This PR is made to discuss my approach to add cookie support and to resolve the issue in near future.

## Related Issues

- Related Issue #465

### Checklist
- Yes I have gone through the [contributing guide](https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md)
- Yes I have updated my branch and synced it with project `main` branch before making this PR
- Yes I have run the tests (`flutter test`) and all tests are passing

## Added/updated tests?

- [ ] No, and this is why:feat yet to be accomplished after that only i can add the tests

### DIFF
--- a/lib/consts.dart
+++ b/lib/consts.dart
@@ -47,6 +47,8 @@ const kTextButtonMinWidth = 44.0;
 
 const kRandMax = 100000;
 
+const kRowRequestTop = 30.0;
+
 const kSuggestionsMenuWidth = 300.0;
 const kSuggestionsMenuMaxHeight = 200.0;
 
@@ -468,3 +470,5 @@ const kMsgNoContent = "No content";
 const kMsgUnknowContentType = "Unknown Response Content-Type";
 // Workspace Selector
 const kMsgSelectWorkspace = "Create your workspace";
+
+const kLabelCookie = "Cookie";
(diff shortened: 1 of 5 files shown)


Clasherzz:
@ashitaprasad  I have put up a button to access cookies and also added a getter to which extracts cookies from http response.  
The cookies tab would also make perfect sense right next to response body and headers. I would like to know if there is anyother options



Screenshot 2024-11-11 224731


I would like to furthur understand how the cookies needs to be stored and reused. From what i have heard the cookies needs to be stored and reused  next time. So should cookies needs to be stored according to domain name or something. The current getter can only have  cookies=cookiesvalue formatted cookies as its the only thing that needs to be send again should the other informations be also stored (like path,timestamp). Also  how should the reuse of cookie from response should work. Like would the cookie header be added  to request upon clicking a cookie extracted from response header.

ashitaprasad:
@Clasherzz Pick up an easier issue to work on .. this feature will require much discussion before any PR is drafted.

Clasherzz:
@ashitaprasad  could i maybe still give it a try  . I can open up a discussion sharing my initial approach.

Issue #490, opened 2024-11-04
Paste not working on iOS and desktop (Right Click -> Paste)


salarshad:
#### Describe the bug/problem
Trying to paste in the endpoint URL or the URL Params or Headers, simply fails without displaying any error. Pasting in the Body text box works as expected 

#### Steps to Reproduce the bug/problem
1. Copy any text from outside the app
2. Long press with the cursor in the endpoint field, or the Params (either the key or the value), or the Headers, and press Paste 

#### Expected behavior
The copied text should be pasted into the text field under focus 

#### Device Info
 - OS: iOS
 - Version: iOS 18.1

#### Flutter Doctor
Please run the `flutter doctor -v` command and provide the details below:

Not Applicable

ashitaprasad:
@salarshad Thanks for pointing out the issue. We have recreated the issue at our end.

ashitaprasad:
Updated title as per the findings in https://github.com/foss42/apidash/pull/491

ashitaprasad:
After testing on latest builds, this issue is no longer present.

PR #498, opened 2024-11-13
Fix: Paste not working on iOS and desktop (Right Click -> Paste)

Lakshya0257:
## PR Description
In iOS and Desktop, right click paste (double tap paste) was not working, along with no error was fired during the event.

Expected behaviour
The copied text should be pasted into the text field on right click paste event.

## Related Issues
#490 

## Reason
ExtendedTextField used for highlighting the variables in the fields like url, header value, param value etc, but the widget does not have a default function like right click paste. Normal TextField works as intended but the text highlighting of variables to red like {{url}}/endpoint is not possible. 

## Added/updated tests?

- Edited and extended the default TextEditingController class for highlighting the variables, and enabling the default Ctrl+V along with right click paste option.
- Changed the ExtendedTextField to default TextField for fixing the issue.

## Output

https://github.com/user-attachments/assets/a66f3126-6eb7-4381-97ca-45b6ca16000d

### DIFF
--- a/lib/screens/common_widgets/env_text_controller.dart
+++ b/lib/screens/common_widgets/env_text_controller.dart
@@ -1,0 +1,49 @@
+import 'package:flutter/material.dart';
+
+class VarTextEditingController extends TextEditingController {
+  final TextStyle normalStyle;
+  final TextStyle highlightStyle;
+
+  VarTextEditingController({
+    required this.normalStyle,
+    required this.highlightStyle,
+    super.text,
+  });
+
+  @override
+  TextSpan buildTextSpan({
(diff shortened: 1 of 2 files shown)


Lakshya0257:
@ashitaprasad can you please review it.

ashitaprasad:
Environment variable colouring is incorrect.
Kindly use the environment variables feature to understand how the colouring works.

Lakshya0257:
Sorry but I don't quite understand, as I think the environment variables ( {{...}} ) should have red color.

ashitaprasad:
Red color is only for those environments variables whose values are not defined.

PR #504, opened 2024-11-24
Implemented FormData parsing in curl_parser and added tests for it.

### LINKED ISSUE #503: [curl_parser] Improvement to the way curl parsing handles form-data
The current implementation of the Curl class does not correctly handle form data (-F flag) in cURL commands. Specifically, multipart form data is not being parsed or included in the resulting Curl object, which limits support for requests requiring file uploads or complex form submissions. This issue affects the ability to fully represent cURL commands in Dart

I was thinking of improving the parsing by allowing multiple -F flags and then returning an additional formFields Map<String,dynamic> but this will have an issue. As you might know sending out lists in formdata we are required to do it like 

image

image

So if we like to parse the curl it will be having curl .... -F "arr=ele1" -F "arr=ele2"

which should result in a map of arr = [ele1,ele2]

Hence we wouldn't be able to return formData as a Map because of the non-unique key therefore I think a list of tuples will be better or a list of formdata class objects

This can be achieved easily I am just writing up my approach.

Suggestions?

WrathOP:
## PR Description

_Add your description_

## Related Issues

- Closes #503 

### Checklist

## Added/updated tests?

- [x] Yes

- [ ] Windows
- [x] macOS
- [ ] Linux

### DIFF
--- a/packages/curl_parser/test/curl_parser_test.dart
+++ b/packages/curl_parser/test/curl_parser_test.dart
@@ -1,5 +1,4 @@
 import 'dart:io';
-
 import 'package:apidash_core/consts.dart';
 import 'package:apidash_core/models/form_data_model.dart';
 import 'package:curl_parser/curl_parser.dart';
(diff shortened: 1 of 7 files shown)


ashitaprasad:
@WrathOP can you segregate the tests into three files:
- `curl_model_test.dart` -> Where you test curl model files are getting creating properly
- `curl_parser_test.dart` -> where curl strings are parsed and checked if correct curl model is being obtained.
- `dart_to_curl_test.dart` -> where curl models are converted into curl string and checked if the correct curl command string is obtained.

WrathOP:
Ok sure will do that.

WrathOP:
@ashitaprasad I have split up the files.

packages/curl_parser/test/dart_to_curl_test.dart
```diff
@@ -0,0 +1,218 @@
+import 'package:apidash_core/consts.dart';
```
ashitaprasad, line 1: Have you manually checked if each and every generated string in this file is correct?

packages/curl_parser/test/curl_parser_test.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:io';
```
ashitaprasad, line 1: Have you manually checked if each and every generated Curl model in this file is correct?

packages/curl_parser/test/dart_to_curl_test.dart
WrathOP: Thanks for catching that, they were not a valid curl to begin with fixed that. BTW I am using postman as my curl test I hope that it should be good.

ashitaprasad:
Thanks for the PR @WrathOP 🚀

PR #509, opened 2024-12-11
Fix: Deletion of Global Environment variable causes null error

sai-phaneesh:
## PR Description
Disabled the delete button when Global environment variable is selected.

### Checklist

## Added/updated tests?
- [ ] Yes
- [x] No, and this is why: The change only involves disabling a button, which prevents the delete action. The functionality does not introduce or modify logic that requires testing.

- [ ] Windows
- [x] macOS
- [ ] Linux

## UI Update
api-dash-bug-fix


ashitaprasad:
Thank you for sending the PR. 
LGTM 🚀

PR #512, opened 2024-12-19
feat:Bypass SSL Verification while making requests

### LINKED ISSUE #486: Add ability to send https requests without SSL verification
Unable to send https requests to servers that don't have certificates, such as local IP addresses

Add function for curl request to be sent with the -k option, ignoring the certificate verification step

Clasherzz:
## PR Description
This PR adds SSL bypasing feature to the http client . Using a callbackMethod

## Related Issues

- Closes #486 

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: Felt its better to add test for createHttpClientWithNoSSL() after getting approval and also had some doubts on how to approach it

- [x] Windows
- [ ] macOS
- [ ] Linux


Clasherzz:
@ashitaprasad Ready for Review

packages/apidash_core/lib/utils/http_request_utils.dart
```diff
@@ -1,6 +1,18 @@
...
 import 'package:seed/seed.dart';
 
+
+http.Client createHttpClientWithNoSSL() {
+  var ioClient = HttpClient()
+    ..badCertificateCallback =
+        (X509Certificate cert, String host, int port) => true;
+  return IOClient(ioClient);
+}
+
```
ashitaprasad, line 15: Will not work in web.
Check out this file - https://github.com/dart-lang/http/blob/master/pkgs/http/lib/src/client.dart

Clasherzz:
@ashitaprasad yes the flutter web framework communicates with browser for ssl verification. There is no way to directly disable or enable it using anykind of code as it posses a security risk. 

I tried running the browser using --ignore-certificate-errors and enabling insecure origin in the developer options of a browser. But it doesn't apply to  XMLHttpRequest API which interacts with flutter only to request made by browser. In short this feauture can't be used on browsers due to security limitations set around by browsers. I should have foreseen this, the best i can do is disable this feauture for web .
 
Do tell me if there is anyother way to make this happen for web (there is a workround method of implementing a middle man and that makes the request through its own client, which is not running on browser and returns it to the actual application running localhost,but i feel like that is not something the organization would want to deal with).

No matter what approach is used eventually we would run into the same issue of browser security .

lib/consts.dart
```diff
@@ -14,6 +14,7 @@ const kAssetSendingLottie = "assets/sending.json";
 const kAssetSavingLottie = "assets/saving.json";
 const kAssetSavedLottie = "assets/completed.json";
 
+final kIsBrowser = kIsWeb;
```
ashitaprasad, line 17: kIsBrowser is a redundancy, remove it & use kIsWeb only.

ashitaprasad:
> Hi @ashitaprasad,
> 
> I initially selected IOClient for its flexibility, particularly with features like SSL bypassing. However, as you pointed out, this approach doesn’t work as expected on the web due to the way the Flutter web framework interacts with the browser for SSL verification. Unfortunately, browsers enforce strict security rules around SSL and HTTPS requests, and there’s no way to bypass or directly alter this behavior through code, as it poses a security risk.
> 
> I attempted to work around this by launching the browser with the --ignore-certificate-errors flag and enabling insecure origins in the developer options. However, I found that this doesn't apply to the XMLHttpRequest API used by Flutter for web, which handles requests made by the browser. In short, due to browser security limitations, it's not possible to disable SSL verification or bypass it directly for web-based applications.
> 
> At this point, it seems like the best course of action is to disable this feature specifically for the web platform, as enabling SSL bypass in browsers isn’t a feasible solution.
> 
> There is a workaround involving a middleman approach, where requests are made through a proxy (outside the browser environment) and then returned to the Flutter web application. However, I understand that this might not be a preferred approach due to the added complexity and infrastructure it would require just for a small feauture.Not to mention we might have to do something for both proxy and web application to be running upon building it(using flutter build).
> 
> Ultimately, it seems we’re constrained by browser security policies, which make it challenging to implement SSL bypass in this case. I hope this provides some clarity on the limitations for web support, but I’m happy to discuss further if you have any alternative suggestions

Yes it will be a good idea to keep it disabled for web.

PR #524, opened 2025-01-05
feat: add oauth

### LINKED ISSUE #481: Implement OAuth 2.0 authentication
I want to send requests to services that use OAuth 2.0. Right now I have to manually acquire an authorization token and copy/paste it into APIDash. The tokens expire usually in an hour, so I have to do this multiple times a day, not only separately for every service, but even for every individual request I have in the library.

APIDash should automatically open a browser session/popup and let me sign in using OAuth 2.0's "hybrid", "implicit" flows. It should also allow me to obtain a token using the "client credentials" flow, which doesn't require a browser session (this is way easier to implement and would already be a huge step forward).

abhinavs1920:
## PR Description

This pull request implements the core functionalities of the OAuth2 feature, specifically focusing on configuration saving and changes. The implementation utilizes the official Dart OAuth2 package and uses GitHub as the OAuth provider for illustration purposes. The Authorization Code flow type is currently implemented.

## Related Issues

- Closes #481

### Checklist
- [x] I have updated my branch and synced it with the project main branch before making this PR
- [x] I am using the latest Flutter stable branch (run flutter upgrade and verify)
- [x] I have run the tests (flutter test) and all tests are passing

## Added/updated tests?

- [x] Yes

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/codegen/codegen_utils.dart
+++ b/lib/codegen/codegen_utils.dart
@@ -1,3 +1,6 @@
+import 'package:apidash_core/models/http_request_model.dart';
+import 'package:seed/models/name_value_model.dart';
+
 String jsonToPyDict(String jsonString) {
   Map<String, String> replaceWithMap = {
     "null": "None",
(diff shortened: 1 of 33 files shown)


ashitaprasad:
This PR contains a lot of unnecessary changes and shows lack of knowledge of existing code.
I recommend you to go through the entire API Dash codebase first to develop a better understanding.

abhinavs1920:
@ashitaprasad Yep, actually that was my brute force approach, I just wanted to know if the flow I implemented currently is right or not. I will make a new PR with clean and optimized code with the current flow I implemented.

PR #552, opened 2025-01-25
added check for new versions and automatically prompt the user

### LINKED ISSUE #373: [feat] in-app update check
Implement feature to automate checking for app updates on launch as mentioned below.

On launch, we can check if a new version of API Dash is available by matching it with GitHub repository. This can be done by
- Maintaining a `version` file in the root directory
- By using GitHub API to find latest release.

image

Then an user prompt can be made to notify user that a new version is available for download. User may have an option to download now, or skip this update.

- If user skips,
  - Save this action in shared preferences and not alert user for current stable version update.

- If user updates, 
  - Scenario 1 (Automatic Update)
    - Use GitHub API to download latest artifact according to user's platform
    - Perform an update (Approach to be discussed)
  - Scenario 2 (Manual Update)
    - Redirect user to API Dash's release tab (https://github.com/foss42/apidash/releases)
    - Let user perform the update 

This feature isn't a priority but a good to have, and may be merged after GSoC contribution period to avoid any conflicts.

abhinavs1920:
## PR Description

This PR implements the in-app update check feature for APIDash. The feature checks for the latest version of the application using the GitHub API and notifies users if an update is available. It supports update handling for Windows, Linux, Android, iOS and macOS platforms.

## Related Issues

- Closes #373 

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: 
I would add the tests when the current approach to solve the issue is finalized.

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,3 +1,4 @@
+import 'package:apidash/providers/update_provider.dart';
 import 'package:apidash_design_system/apidash_design_system.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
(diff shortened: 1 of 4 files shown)


ashitaprasad:
App update should be silent and optional .. maybe a dart isolate that spawns, checks and does not block the main thread when the user starts the app. Accessible via settings and a red dot on settings icon button can indicate an update is available.

PR #554, opened 2025-01-27
Added url encoded format for sending the data in request

### LINKED ISSUE #337: Support for application/x-www-url-encoded Content-Type as body type formdata currently only supports multipart/form-data
**Describe the bug/problem**
Currently whenever the formdata is selected as request body, the `Content-Type` header is set to `multipart/form-data` but this should be dependent upon whether there is a file being uploaded too or else it can be set to the default `application/x-www-form-urlencoded`. 

**Relevant References** 
[stackoverflow: application/x-www-form-urlencoded or multipart/form-data?](https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data)
[baeldung: Difference Between form-data, x-www-form-urlencoded and raw in Postman](https://www.baeldung.com/postman-form-data-raw-x-www-form-urlencoded)

Select POST request in the request method, select formdata as the request body, add only text fields as fields, and then we can either chck the codegen for languages where Content-Type is printed, or else we can print `requestModel.requestBodyContentType.header` inside codegen files to check the request header.

image

**Expected behavior**
If there aren't any files uploaded i.e. all the fields are of text type and selected type is formdata, then the `Content-Type` should be set to `application/x-www-url-encoded`, and if there is at least one file in the form data, then the `Content-Type` should be `multipart/form-data`. As the default `Content-Type` for HTML form submission is `application/x-www-form-urlencoded`, hence it is worth introducing in API dash.

Reference: [Mozilla Documentation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data)
image

Fixing this should start from introducing a new sub-type inside `consts.dart` file and then creating a new `ContentType` for the new type and setting up the Content Type based on whether there are any files associated with the request or not. Below are some code snippets from `consts.dart` which are to be modified to start working on this issue.

image
image

 - OS: Debian 12 WSL2 on Windows 11

abhinavs1920:
## PR Description

This PR implements a toggle switch feature in the form-data section of the API Dash application. The toggle allows users to switch between application/x-www-form-urlencoded and multipart/form-data content types. The feature automatically selects the appropriate content type based on whether files are included in the form data, enhancing user experience and ensuring correct data submission.

## Related Issues

- Closes #337 

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: I am currently waiting for my approach to get completely verified.

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/codegen/dart/dio.dart
+++ b/lib/codegen/dart/dio.dart
@@ -84,6 +84,14 @@ class DartDioCodeGen {
         // when add new type of [ContentType], need update [dataExp].
         case ContentType.formdata:
           dataExp = declareFinal('data').assign(refer('dio.FormData()'));
+        case ContentType.urlencoded:
+          // Convert form data to URL-encoded format
+          final urlEncodedData = formData.map((item) => 
+            '${Uri.encodeComponent(item['name'] ?? '')}'
+            '='
+            '${Uri.encodeComponent(item['value'] ?? '')}'
+          ).join('&');
+          dataExp = declareFinal('data').assign(literalString(urlEncodedData));
       }
     }
     final responseExp = declareFinal('response').assign(InvokeExpression.newOf(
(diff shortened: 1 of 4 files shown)


ashitaprasad:
The current design and implementation are both incorrect. Please discuss your approach on the issue first along with the design, then you should send the PR after it is finalized.

PR #567, opened 2025-02-11
Initial working xwwwurlencoded body type

### LINKED ISSUE #337: Support for application/x-www-url-encoded Content-Type as body type formdata currently only supports multipart/form-data

WrathOP:
## PR Description

Implemented a basic POC for the xwwwurlencoded body content by adding an option to send it seperate from already implemented formdata. I think doing it this way is better as the discussed way of sending by default url encoded and sending multipart when a file is attached can lead to confusions and also if the user specificly wants to send a multipart. If we are adding a toggle for these 2 it will create a sort of divide in the type of the body type where you will have a different variable of sorts to know what kind of body this is. Anyways open to suggestions.

## Related Issues

https://github.com/user-attachments/assets/d3c3dd7a-77ce-4bc7-81fd-7cf8c2d5b827

- Closes #337 

### Checklist

## Added/updated tests?
I'll add tests in further development after the POC is approved

- [ ] Yes
- [x] No, and this is why:I'll add tests in further development after the POC is approved

- [ ] Windows
- [x] macOS
- [ ] Linux

### DIFF
--- a/lib/codegen/dart/dio.dart
+++ b/lib/codegen/dart/dio.dart
@@ -84,6 +84,8 @@ class DartDioCodeGen {
         // when add new type of [ContentType], need update [dataExp].
         case ContentType.formdata:
           dataExp = declareFinal('data').assign(refer('dio.FormData()'));
+        case ContentType.xwwwformurlencoded:
+          dataExp = declareFinal('data').assign(refer('dio.FormData()'));
       }
     }
     final responseExp = declareFinal('response').assign(InvokeExpression.newOf(
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -49,6 +49,7 @@ Future<bool> initApp(
     if (openBoxesStatus) {
       await autoClearHistory(settingsModel: settingsModel);
     }
+    debugPrint("autoClearHistory done");
     return openBoxesStatus;
   } catch (e) {
     debugPrint("initApp failed due to $e");
--- a/lib/screens/history/history_widgets/his_request_pane.dart
+++ b/lib/screens/history/history_widgets/his_request_pane.dart
@@ -133,10 +133,11 @@ class HisRequestBody extends ConsumerWidget {
             kVSpacer5,
             Expanded(
               child: switch (contentType) {
-                ContentType.formdata => Padding(
-                    padding: kPh4,
-                    child: RequestFormDataTable(
-                        rows: requestModel?.formData ?? [])),
+                (ContentType.formdata || ContentType.xwwwformurlencoded) =>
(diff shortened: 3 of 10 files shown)


ashitaprasad:
The second option provided in this [comment](https://github.com/foss42/apidash/issues/337#issuecomment-2057664500) will be more user friendly

WrathOP:
https://github.com/user-attachments/assets/2f2e578f-ea75-41f9-af05-a35a3f8589a6

Here is the POC for this approach. Just need to figure out where to put this switch as it looks so out of place here.

ashitaprasad:
> Screen.Recording.2025-02-21.at.2.17.36.PM.mov 
> Here is the POC for this approach. Just need to figure out where to put this switch as it looks so out of place here.

The POC looks good @WrathOP 👍

ashitaprasad:
You can cleanup the code and let me know when it is ready for review.

Issue #551, opened 2025-01-25
Clear History via isolate


ashitaprasad:
Currently, history is cleared in the main thread. Optimize the code to clear it using isolates.

abhinavs1920:
To optimize clearing history data using isolates, we need to focus on offloading this task from the main thread to improve performance and responsiveness.

The first step is identifying where the history clearing logic currently resides, likely within a model or service related to history management. Once located, we should extract this logic into a standalone function that can be executed in an isolate.

Next, we utilize Dart's isolate capabilities, specifically the Isolate.spawn function, to run this clearing function in a separate thread. This involves setting up a ReceivePort to facilitate communication between the main thread and the isolate, allowing us to receive confirmation once the clearing operation is complete. The data necessary for the clearing process, such as references to the history storage or criteria for identifying outdated entries, should be prepared and passed to the isolate. 

After implementing the isolate-based clearing function, we integrate it into the existing history management logic, replacing any direct clearing operations on the main thread. This ensures that the application remains responsive, even when handling large volumes of history data. 

Please let me know if the approach is good enough.

ashitaprasad:
@abhinavs1920  You can share across a draft PR highlighting your approach.

ashitaprasad:
Thnaks for your proposal @Vishwa-Karthik

deepak0x:
Is this still open? The earlier PRs (#575, #577, #604) were all closed and nothing seems to have landed. I would like to pick this up — clearing expired history records off the main thread via `compute`/isolate, with batched deletes for large datasets. @ashitaprasad @animator could you assign it to me if it is still available?

PR #577, opened 2025-02-19
[Proposal] Issue #551 - Clear History Via Isolates

Vishwa-Karthik:
## Clear History Via Isolates for Open Issue #551

Attached markdown file with the proposal and a few questions to streamline the requirements.

cc: @ashitaprasad


ashitaprasad:
Instead of sending the proposal (kindly specify if it is for GSoC, FOSSHACK or a regular contribution?) as a PR.
You can put your thoughts in the form of a comment on the relevant issue or raise a new issue in case there is no relevant issue.

PR #578, opened 2025-02-19
[Proposal] Enhancement - JSON Editor View - Proposal

Vishwa-Karthik:
## JSON Editor View

When JSON type is selected from BODY for Mutation Protocols, how about we incorporate JSON Editor View, which has the capabilities mentioned below.

## Features 
+ Syntax Highlighting
+ JSON Validator
+ JSON Beautification
+ Object Indentation
+ Automatic Bracket Completer
+ Implement Line Number
+ Show Error Line
+ Allow comments? (should we include this?)

cc: @ashitaprasad

### DIFF
--- a/json_editor_view_proposal.md
+++ b/json_editor_view_proposal.md
@@ -1,0 +1,34 @@
+# Proposal: Implementing a JSON Editor View for ApiDash
+
+I encountered this missing functionality when compared with other softwares in the market where JSON Editor 
+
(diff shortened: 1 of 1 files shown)


ashitaprasad:
Instead of sending the proposal (kindly specify if it is for GSoC, FOSSHACK or a regular contribution?) as a PR, you can put your thoughts in the form of a comment on the relevant issue or raise a new issue in case there is no relevant issue.

PR #588, opened 2025-02-21
feat:graphql_variables

### LINKED ISSUE #576: GraphQL Variables
Finding it hard to use the graphql feauture since the query needs to be edited again and again for minor changes in values.

Introduce a section for graphql variables .

Clasherzz:
## PR Description
Trying to add graphQL feauture

## Related Issues

- Closes #576

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: will like to know whether the initial approach is correct

- [x] Windows
- [ ] macOS
- [ ] Linux


Clasherzz:
image
image

@ashitaprasad  Haven't updated the tests . Would like to know if the approach is correct and if PR has some issues , first ?

ashitaprasad:
In GraphQL, the standard practice is to provide variables in the form of JSON. So, a simple JSON text editor (like that for JSON content-type request body) should be preferred.

Clasherzz:
> In GraphQL, the standard practice is to provide variables in the form of JSON. So, a simple JSON text editor (like that for JSON content-type request body) should be preferred.
> 
> Instead of parsing form inputs, handling type conversions and rebuilding the json, this will be more efficient.
> 
> And, it will also pave way for other graphQL features like filters & mutation.
> 
> Ref:
> 
> * https://www.contentful.com/blog/how-to-use-graphql-variables/
> * https://dgraph.io/docs/query-language/graphql-variables/
> * https://hygraph.com/docs/api-reference/content-api/variables

I see i wasnt aware of that . Would look into it in detail

Issue #122, opened 2024-02-22
Importing Requests from HAR file


ashitaprasad:
**Tell us about the task you want to perform and are unable to do so because the feature is not available**
Add support for importing requests from the HAR file which can be exported from Chrome. The developed code should be able to parse the HAR files, after processing let user choose the requests he wants to add via a dialog and create the corresponding requests in the API Dash collection.

opxdelwin:
Hey @ashitaprasad 

I've been working on this implementation, for sure it'd be a starting point for multiple type of imports (such as POSTMAN, etc.).
Please do have a look into the PR: #158 

Thanks !

luxmikant:
hello maintainer and fellow contributors i would like to contribute to this project .

ashitaprasad:
You can go ahead and send a draft PR @luxmikant

luxmikant:
hey actually iam geting some error that uable to prase file_name.har how i can fix that

ashitaprasad:
> hey actually iam geting some error that uable to prase file_name.har how i can fix that

How are you parsing the HAR file?

ayaan-md-blr:
@ashitaprasad I would like to contribute to this PR. Have also submitted an idea doc for the AI UI visualizations.

PR #598, opened 2025-02-22
intial commit of importing HAR file  option

luxmikant:
## PR Description

Draft PR for review.

## Related Issues

importing Requests from HAR file #122

### Checklist
- [* ] I have gone through the [contributing guide](https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md)
- [* ] I have updated my branch and synced it with project `main` branch before making this PR
- [* ] I am using the latest Flutter stable branch (run `flutter upgrade` and verify)
- [ *] I have run the tests (`flutter test`) and all tests are passing

## Added/updated tests?

- [ *] Yes

- [ *] Windows
- [ ] macOS
- [ ] Linux


ashitaprasad:
@luxmikant You have to add test cases like we have for Postman importer.

luxmikant:
hey can you please help with testing i am new to testing 
how can i put test case and how to run them 

The HAR parsing is implemented using two classes: Har and HarEntry.

Har.fromJson Constructor:

This takes a JSON map representing the full HAR file.
It first validates that the top-level JSON contains a 'log' object.
It then checks that this log object contains an 'entries' array.
The constructor filters the entries to include only those with a non-null 'request' property.
For each entry, it calls HarEntry.fromJson passing the inner JSON of the 'request' key.
Finally, it returns a Har instance with these parsed HarEntry objects.
HarEntry.fromJson Constructor:

This constructor is responsible for parsing a single request entry.
It extracts the required properties: method and url.
For headers, it reads the list found under 'headers'. It iterates over each header and:
Checks if both 'name' and 'value' exist.
Filters out any headers where the name starts with ":" (for example, ":authority"), which are considered invalid HTTP header names.
It also checks for an optional body property by looking at the 'postData' object. If present and if it contains the key 'text', the associated text is used as the body.
Finally, it creates and returns a HarEntry with the parsed values.

ashitaprasad:
Like I said in my comment you need to take a look at the test files for postman collection importer to understand how it is written.
Also, https://github.com/foss42/apidash/blob/main/doc/dev_guide/testing.md will help

ashitaprasad:
Apart from testing, the structure of package is also incorrect. I recommend you to take a look at other importer packages like postman, curl or insomnia and re-work on the PR.

PR #631, opened 2025-03-02
Fix: Prevent automatic charset=utf-8 addition to Content-Type headers

### LINKED ISSUE #630: Unable to override Content-Type header
#### Describe the bug/problem

Create a new POST request as shown below with JSON body

Image

If you do not specify any content-type in header, the client automatically sets it to `application/json; charset=utf-8`.

In the headers, you can specify the `Content-Type` as `application/json` to override the default and send request. When you inspect headers in response pane you can observe that the Request header that was sent for `Content-Type` was  `application/json; charset=utf-8` instead of `application/json`

Image

`charset=utf-8' is automatically added into 'Content-Type', and we are unable to remove it as for some APIs, this header string can cause rejection.

Similarly, if you set `text/csv` as Content-Type in Request and send the response. Again `charset=utf-8' is being automatically added by `http` library.

Image

The objective is that user should have full control over the `Content-Type` header that is being sent to the server.

Relevant issues:
- https://github.com/dart-lang/http/issues/1438
- https://github.com/dart-lang/http/issues/184

AffanShaikhsurab:
# PR Description

This change prevents the HTTP package from automatically appending 'charset=utf-8' to Content-Type headers by:

1. Using low-level http.Request instead of high-level client methods
2. Setting request.bodyBytes directly with UTF-8 encoded content 
3. Preserving exact Content-Type headers as specified by users

This ensures users have full control over Content-Type headers sent to servers, addressing issues where some APIs reject requests with charset parameters.

## Related Issues

- Closes #630

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: This fix resolves an urgent issue with Content-Type headers. Proper testing requires mocking HTTP client responses which will be implemented in a separate pr to focus on comprehensive test coverage for HTTP service functionality. We can add proper testing in this pr also after proper discussion 

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/packages/apidash_core/lib/services/http_service.dart
+++ b/packages/apidash_core/lib/services/http_service.dart
@@ -71,37 +71,31 @@ Future<(HttpResponse?, Duration?, String?)> sendHttpRequest(
               }
             }
             http.StreamedResponse multiPartResponse =
-                await multiPartRequest.send();
+                await client.send(multiPartRequest);
             stopwatch.stop();
             http.Response convertedMultiPartResponse =
                 await convertStreamedResponse(multiPartResponse);
             return (convertedMultiPartResponse, stopwatch.elapsed, null);
           }
         }
-        switch (requestModel.method) {
-          case HTTPVerb.get:
-            response = await client.get(requestUrl, headers: headers);
-            break;
-          case HTTPVerb.head:
-            response = await client.head(requestUrl, headers: headers);
-            break;
-          case HTTPVerb.post:
-            response =
-                await client.post(requestUrl, headers: headers, body: body);
-            break;
-          case HTTPVerb.put:
-            response =
-                await client.put(requestUrl, headers: headers, body: body);
-            break;
-          case HTTPVerb.patch:
-            response =
-                await client.patch(requestUrl, headers: headers, body: body);
-            break;
-          case HTTPVerb.delete:
-            response =
-                await client.delete(requestUrl, headers: headers, body: body);
-            break;
+        
+        // Replace the switch statement with manual Request creation
+        final request = http.Request(
+          requestModel.method.name.toUpperCase(),
+          requestUrl,
+        );
+        
+        // Set headers
+        request.headers.addAll(headers);
+        
+        // Set body if needed
+        if (body != null && kMethodsWithBody.contains(requestModel.method)) {
+          request.bodyBytes = utf8.encode(body);
         }
+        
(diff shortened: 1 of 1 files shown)


packages/apidash_core/lib/services/http_service.dart
```diff
@@ -1,17 +1,198 @@
...
+Encoding getEncodingFromCharset(String? charsetName) {
+  if (charsetName == null) {
+    return utf8;
+  }
+  
+  // Normalize to lowercase for case-insensitive matching
+  final normalizedName = charsetName.toLowerCase().trim();
+  
+  // Direct codec mappings - built-in encodings first
+  switch (normalizedName) {
```
ashitaprasad, line 28: What is the source of this codec map and names?

AffanShaikhsurab: https://pub.dev/documentation/charset/latest/charset/

AffanShaikhsurab: I think we don’t need this much but I have just added it we can remove them also

ashitaprasad: @AffanShaikhsurab Where in the link are the strings `iso646-us` & `iso_8859-1:1987` that was written below in the case matching?
Kindly do not beat around the bush and tell exactly how you got all the string cases.

AffanShaikhsurab: @ashitaprasad  I initially thought that the ISO encoding was present in the list of supported encodings, but upon further inspection, I realized it was not included in the updated code. I have now ensured that only the encoding types available in the dart:convert package and Flutter’s Charset are used.

I have a question: should we first encode it as UTF-8 and then pass it as text/csv in the content headers , is this the expected behaviour ?

Apologies for the earlier oversight—I was working on multiple issues simultaneously, which led to an incomplete implementation. I will ensure that I take on fewer tasks at a time and implement them more carefully moving forward.

animator: @AffanShaikhsurab You still haven't given a direct answer to my question. 
The link you initially shared did not contain the encoding strings you had written in the code when I clearly asked you for the source.
I am asking for one last time - What was the source of the long list of encoding strings that you had provided initially in the code.
I am just expecting a single line answer with the URL Link.

PR #662, opened 2025-03-13
Env variable from OS

### LINKED ISSUE #600: Reading environment variables directly from OS environment

It would be great if API Dash has the ability to read environment variables from the operating system. This way it can access security credentials (like a password) and other system level variables.

abhinavs1920:
## PR Description

This PR adds support for fetching environment variables directly from the OS in API Dash. Users can now specify in the environment manager that a variable should be pulled from the system's environment variables instead of being stored in API Dash. The feature enhances security by allowing users to reference sensitive credentials (e.g., API keys, tokens, passwords) without exposing them in the app.

## Related Issues

- Closes #600 

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why:  Will include the tests as soon as my approach will be accepted

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/screens/envvar/editor_pane/variables_pane.dart
+++ b/lib/screens/envvar/editor_pane/variables_pane.dart
@@ -70,6 +70,11 @@ class EditEnvironmentVariablesState
         label: Text("Variable value"),
       ),
       DataColumn2(
+        label: Text('OS'),
+        fixedWidth: 50,
+        tooltip: 'Use OS environment variable',
+      ),
+      DataColumn2(
         label: Text(''),
         fixedWidth: 32,
       ),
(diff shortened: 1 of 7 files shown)


abhinavs1920:
@ashitaprasad 
Please check!
I am excited to get corrections and work on them.
Thanks.

ashitaprasad:
Resolve the code conflicts with the latest main branch.

animator:
Update the code as per latest main branch.
Also, add tests.

abhinavs1920:
> Update the code as per latest main branch.
> Also, add tests.

Sure, will do it.

PR #750, opened 2025-03-31
Fix issue #630: Unable to override Content-Type header charset

### LINKED ISSUE #630: Unable to override Content-Type header

Vastav1812:
# Fix: Prevent automatic charset=utf-8 addition to Content-Type headers

## Issue
When sending requests with Content-Type headers like "application/json", the Dart HTTP library automatically appends "charset=utf-8" to the header value. Some APIs reject requests with this charset specification, but users currently have no way to prevent this behavior.

Fixes #630 by giving users control over Content-Type headers.

## Changes
- Added `useRawContentType` boolean field to `HttpRequestModel`
- Modified HTTP service to respect this setting when setting Content-Type headers
- Added UI checkbox in Headers section labeled "Prevent charset in Content-Type"
- Added tooltip and info dialog explaining the purpose of this option
- Added application-wide setting to control default behavior for all requests
- Added Content-Type header preview in Response Headers section

## Implementation Details
- When `useRawContentType` is true, Content-Type headers are sent without charset
- When false (default), the standard behavior is maintained for compatibility
- New requests inherit the application-wide setting

## Testing Instructions
1. Create a POST request with JSON body
2. Observe that by default, Content-Type includes charset=utf-8
3. Enable "Prevent charset in Content-Type" in Headers section
4. Send request and verify Content-Type header is sent without charset
5. Try with application-wide setting enabled to verify it applies to new requests

## Related Issues
- dart-lang/http#184
- dart-lang/http#1438

### DIFF
--- a/packages/apidash_core/lib/models/http_request_model.dart
+++ b/packages/apidash_core/lib/models/http_request_model.dart
@@ -27,6 +27,7 @@ class HttpRequestModel with _$HttpRequestModel {
     @Default(ContentType.json) ContentType bodyContentType,
     String? body,
     String? query,
+    @Default(false) bool useRawContentType,
     List<FormDataModel>? formData,
   }) = _HttpRequestModel;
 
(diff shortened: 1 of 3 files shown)


animator:
Please do not modify the PR template. Also, go through the issue and understand it better as this PR is currently not solving it.

animator:
In case you are participating in GSoC, it is better that your submit your proposal first as the application window has already opened.

PR #839, opened 2025-05-01
Fix Issue #824 [Mobile] Making copying values (like tokens) from JSON previewer easy for end users 

### LINKED ISSUE #824: [Mobile] Making copying values (like tokens) from JSON previewer easy for end users

Related Desktop issue #638 

There should be a one click solution to easily copy values (like token values) to clipboard from JSON previewer for mobile users.
Currently, for desktop users there is a copy button which appears on hover, but no such provision is available for mobile/tablet.

WannaCry016:
## PR Description
This PR adds the feature to copy for mobile users in one click on the key-value pair just like in desktop it shows copy option on hovering. 
So I have replaced the GestureDetector with Listener as this caused issues where taps wouldn’t trigger correctly due to SelectableText or TextFields

https://github.com/user-attachments/assets/5244ce07-99e3-4424-9d23-6b4e51ad0f51

## Related Issues

- Closes #824 

### Checklist
- [X] I have updated my branch and synced it with project `main` branch before making this PR
- [X] I am using the latest Flutter stable branch (run `flutter upgrade` and verify)

## Added/updated tests?
- [ ] Yes
- [X] No, and this is why: Already There

- [X] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/packages/json_explorer/lib/src/json_explorer.dart
+++ b/packages/json_explorer/lib/src/json_explorer.dart
@@ -288,17 +288,19 @@ class JsonAttribute extends StatelessWidget {
         node.highlight(isHighlighted: false);
         node.focus(isFocused: false);
       },
-      child: GestureDetector(
-        behavior: HitTestBehavior.opaque,
-        onTap: hasInteraction
-            ? () {
-                if (valueStyle.onTap != null) {
-                  valueStyle.onTap!.call();
-                } else {
-                  _onTap(context);
-                }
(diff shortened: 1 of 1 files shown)


WannaCry016:
@animator can u review the PR

animator:
Test it for nested responses.
it is not working properly.

WannaCry016:
Ok I will try to fix it

PR #847, opened 2025-05-20
Fix: prevent malformed URL error when parsing IP:PORT by adding defau…

### LINKED ISSUE #841: URL with IP:PORT returns malformed URL error
#### Describe the bug/problem

Below goes as a HTTPS request as expected

Image

Below should also go as a HTTPS request, but throws the following error.

Image

surelyruchi:
## PR Description
This PR fixes an issue where URLs containing IP addresses with ports (e.g., 8.8.8.8:443) were being treated as malformed due to the absence of a URI scheme (http://, https://, etc.).

## Related Issues

- Closes #841

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: This is a small, focused fix , tests will be added later if required based on usage or feedback.

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/packages/apidash_core/lib/consts.dart
+++ b/packages/apidash_core/lib/consts.dart
@@ -30,6 +30,8 @@ final kSupportedUriSchemes =
     SupportedUriSchemes.values.map((i) => i.name).toList();
 const kDefaultUriScheme = SupportedUriSchemes.https;
 final kLocalhostRegex = RegExp(r'^localhost(:\d+)?(/.*)?$');
+final kIPHostRegex =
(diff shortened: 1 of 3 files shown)


sharmaruchi30:
Hey @ashitaprasad @animator, this is a minor change in the code. Could you take a quick look and share your thoughts when you get a chance?

Basically, the issue was URL's like 8.8.8.8:443 without scheme passing through Uri.parse(url) treats it as a string and returns null value which resulted to malformed url and previously default scheme were getting added if the URL matched kLocalhostRegex which only handled localhost.
So I added a generic regex check to add missing URL scheme if missing from such address.

animator:
Sure. We will review the code.

ashitaprasad:
Thanks for the PR.

PR #927, opened 2025-11-27
Fix: TabBar overflow on mode change

MananGandhi1810:
## PR Description
When changing the request mode from HTTP to AI or GraphQL, if the TabBar Index is set to an index higher than the available tabs in the next mode, the app triggers an error. Attaching a video demo of the same. This PR fixes that error and sets the tabbar to index 0 if it overflows. For GraphQL, the tab on index 3 was also being ignored, hence fixed it as well.

https://github.com/user-attachments/assets/56cf9b65-e701-4265-98f7-70606ff6e77f

## Related Issues
None

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: Tests are not needed

- [ ] Windows
- [x] macOS
- [ ] Linux

### DIFF
--- a/lib/providers/collection_providers.dart
+++ b/lib/providers/collection_providers.dart
@@ -250,12 +250,15 @@ class CollectionStateNotifier
       newModel = switch (apiType) {
         APIType.rest || APIType.graphql => currentModel.copyWith(
             apiType: apiType,
+            requestTabIndex: 0,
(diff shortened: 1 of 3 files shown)


animator:
Thanks @MananGandhi1810 for this wonderful catch.
The solution is to reset the tab index to 0 on API Type change.

PR #933, opened 2025-12-11
Fix: Typing or Removing any character in between URL textfield moves …

anshxs:
Fixed the issue 924


animator:
Please follow the contributing guidelines and PR request template.
https://github.com/foss42/apidash/blob/main/CONTRIBUTING.md

PR #949, opened 2025-12-15
Resolves: Issue #826. Display Long URLs in multiple lines

### LINKED ISSUE #826: Textfields should become multiline when tapped on or when the user is editing

URL params or header values can be quite long some times (like bearer token values).
Instead of horizontally scrolling through the field (to verify due to the fear of missing out something) , life will become much easier if the value wraps across multiple lines and one can easily view the full text and edit it.

Example below:

Image

Sanat-Jha:
## PR Description

In case the api url is long and overflowing horizontally, it will wrap in multiline format instead of horizontal scroll when in focus and display one liner when lost focus.

## Related Issues

- Closes #826 

### Checklist

- [x] Windows
- [ ] macOS
- [ ] Linux

When in focus
image

When not in focus
image

### DIFF
--- a/lib/screens/common_widgets/envfield_url.dart
+++ b/lib/screens/common_widgets/envfield_url.dart
@@ -3,7 +3,7 @@ import 'package:flutter/material.dart';
 import 'package:apidash/consts.dart';
 import 'env_trigger_field.dart';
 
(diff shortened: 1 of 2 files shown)


Sanat-Jha:
@animator Please review this PR

PR #1036, opened 2026-01-26
Added configurable default timeout

### LINKED ISSUE #1027: Timeout Support for API Requests
### Tell us about the task you want to perform and are unable to do so because the feature is not available

When testing APIs that return responses after a delay (for example:  
`https://httpstat.us/200?sleep=15000`), API Dash fails with a low-level `ClientException` without providing any control over how long the request should wait or a clear indication that the failure was due to a timeout.

Currently:

- There is no way to configure a request timeout duration.
- Slow but valid APIs may be interpreted as failed requests.
- The error message shown is technical and not clearly associated with a timeout.
- There is no option to cancel an in-progress request.

This makes it difficult to reliably test APIs that legitimately require more time to respond.

---

###  Describe the solution/feature you'd like us to add

Introduce request timeout handling with the following behavior:

- A global default timeout value configurable in settings.
- An optional per-request override when required.
- A clear timeout message such as:  
  > "Request timed out after X seconds."
- Optional ability to cancel an ongoing request.

The default timeout can remain at a sensible value so that existing workflows are not affected unless users choose to modify it.

---

### Any other feedback you would like to provide regarding the site

API Dash already provides a clean and intuitive API testing experience. Adding proper timeout handling would significantly improve reliability and clarity when working with slow or unpredictable APIs.

I would appreciate hearing the maintainers’ thoughts on whether this aligns with the project roadmap before exploring implementation.

hannah-mnq:
## PR Description

This PR adds a **configurable default HTTP request timeout** to the `better_networking` package.

### What’s changed
- Introduced a centralized `kDefaultRequestTimeout` constant
- Wired the timeout through `HttpClientManager`
- Ensured the timeout is respected across:
  - REST requests
  - GraphQL requests
  - Multipart requests
  - Streamed requests

This establishes a safe default timeout while keeping the implementation minimal and non-breaking.  
**Per-request and advanced timeout configuration via the UI will be introduced in a follow-up PR.**

## Related Issues

- Closes #1027

### Checklist

## Added/updated tests?

- [ ] Yes  
- [x] No, and this is why:  
  This change introduces a default configuration and wiring logic without altering existing request behavior or public APIs. Adding tests is better suited once per-request or UI-level timeout configuration is implemented.

- [x] Windows  
- [ ] macOS  
- [ ] Linux

### DIFF
--- a/packages/better_networking/lib/services/http_service.dart
+++ b/packages/better_networking/lib/services/http_service.dart
@@ -31,7 +31,12 @@ Future<(HttpResponse?, Duration?, String?)> sendHttpRequestV1(
   if (httpClientManager.wasRequestCancelled(requestId)) {
     httpClientManager.removeCancelledRequest(requestId);
   }
-  final client = httpClientManager.createClient(requestId, noSSL: noSSL);
+
+  final client = httpClientManager.createClient(
+    requestId,
+    noSSL: noSSL,
+    timeout: kDefaultRequestTimeout,
+  );
 
   HttpRequestModel authenticatedRequestModel = requestModel.copyWith();
 
@@ -92,9 +97,9 @@ Future<(HttpResponse?, Duration?, String?)> sendHttpRequestV1(
                 );
               }
             }
-            http.StreamedResponse multiPartResponse = await client.send(
-              multiPartRequest,
-            );
+            http.StreamedResponse multiPartResponse = await client
+                .send(multiPartRequest)
+                .timeout(kDefaultRequestTimeout);
 
             stopwatch.stop();
             http.Response convertedMultiPartResponse =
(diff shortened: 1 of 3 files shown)


packages/better_networking/lib/services/http_service.dart
```diff
@@ -104,10 +109,14 @@ Future<(HttpResponse?, Duration?, String?)> sendHttpRequestV1(
...
-            response = await client.get(requestUrl, headers: headers);
+            response = await client
+                .get(requestUrl, headers: headers)
+                .timeout(kDefaultRequestTimeout);
             break;
           case HTTPVerb.head:
-            response = await client.head(requestUrl, headers: headers);
+            response = await client
+                .head(requestUrl, headers: headers)
+                .timeout(kDefaultRequestTimeout);
```
animator, line 119: why timeout() here if the timeout defined in client works?

hannah-mnq: > why timeout() here if the timeout defined in client works?

I added .timeout() at the request level mainly as a safety guard so the request Future itself doesn’t hang longer than expected, even if something goes wrong internally.
That said, I agree that if the client-level timeout is already reliable and intended to be the single place where timeouts are handled, then adding .timeout() here is redundant. I’m happy to remove it and keep timeout behavior centralized in the client.

PR #1049, opened 2026-02-01
URL autocomplete using request history

### LINKED ISSUE #1032: Add URL request history with autocomplete in request editor

While working with APIs, the same request URLs are often reused multiple times during development and testing. Currently, API Dash requires retyping or manually copying previously used URLs each time, which can slow down workflows and become repetitive, especially when switching between multiple endpoints.

Add URL request memory with autocomplete in the request editor.  
When a user starts typing in the URL field, API Dash should suggest URLs from previously successful requests. This would help reduce repetitive typing, speed up request creation, and improve the overall developer experience.

The feature can initially be scoped to URL autocomplete only, with the possibility to extend later to other request fields if needed.

This would be a useful developer-experience improvement and aligns well with common patterns in API testing tools.

Bhavesh-More:
## PR Description

This PR adds URL history-based autocomplete to the request URL.

The goal is to make it easier to reuse previously successful request URLs without retyping them, while keeping the behavior simple and predictable.
## Related Issues

- Closes #1032 

## Demo

https://github.com/user-attachments/assets/aa0765b6-78e3-4d9d-a798-6ac710f56f35

### Checklist

## Added/updated tests?

- [x] Yes

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/providers/collection_providers.dart
+++ b/lib/providers/collection_providers.dart
@@ -541,7 +541,11 @@ class CollectionStateNotifier
       if (statusCode >= 200 && statusCode < 300) {
         final rawUrl = requestModel.httpRequestModel?.url ?? '';
         await hiveHandler.saveUrlToHistory(
-            rawUrl, substitutedHttpRequestModel.url);
+            rawUrl,
+            substitutedHttpRequestModel.url,
+            substitutedHttpRequestModel.method.name);
+        // Refresh the URL history provider
+        ref.read(urlHistoryProvider.notifier).refresh();
       }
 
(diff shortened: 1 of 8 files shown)


animator:
A question I have is ... should you show the empty suggestion box if there are no search suggestions?

Bhavesh-More:
I added the “No URL” empty state because I felt it was a better UX to explicitly indicate that there are no suggestions. That said, I’m happy to adjust it if the preferred behavior is to hide the suggestion box when there are no matches.

PR #1061, opened 2026-02-09
Refactor hive & adding autosave functionality 

ShashwatXD:
## PR Description

This PR implements autosave by changing API Dash storage from a Load All / Save All model to a Lazy Load / Granular Save model.

Now the app loads only request metadata at startup, and the full request is loaded only when a user opens it. Any change to a request is saved immediately to Hive instead of saving the entire collection manually.

- Closes #1034 

### Checklist

## Added/updated tests?

- [x] Yes, i have updated "save and load" tests,since the new architecture loads full data lazily.

- [ ] Windows
- [x] macOS
- [ ] Linux


ShashwatXD:
@animator, please review this PR.
I have not removed the persistent Save button in the collection pane yet. Since everything is autosaved now, it no longer does anything, it can be removed in a follow-up if you prefer.
Also, opening the app with existing (old) storage boxes will currently crash because the dataBox type has changed to LazyBox.(as discussed before)

animator:
is this PR now relevant @ShashwatXD ?

ShashwatXD:
> is this PR now relevant @ShashwatXD ?

nope, we can close this one

PR #1070, opened 2026-02-12
feat: add copy value button to JSON previewer

### LINKED ISSUE #638: Make copying just the values from JSON rendered view possible
I do a little login to my api, like this:

Image

and get a token. This token i want on my clipboard obviously. Now I can't select the text to copy it, so i have to use the little copy button, which would be fine if it didnt also copy the json structure around it. When i press the little copy button i get this:

```
{
  "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.eyJwaWQiOiJiOTFmNWZhYS05NWYzLTQ1ZDUtYmM1ZC1kMjcxNjg3NGQ0NmUiLCJleHAiOjE3NDE3ODMwNzAsImNsYWltcyI6bnVsbH0.Okfmc52W6vZzl74-Y5hxLPVefhFatdkzs3pHgEcUc1tG1RTKc2v7quV2nwoEAy-1AVme-Run3698MuX_9yc7QQ"
}
```

Either make the text selectable in the rendered json view, or make the copy button give me the value of the key in the json, not a new json object with just that key.

I like the app, keep it up

kpradyun:
## PR Description

Currently, the copy button in the JSON previewer copies the entire key-value pair as a JSON object (e.g., `{"token": "eyJ..."}`). This PR introduces a second **"Copy Value"** button that extracts only the raw value (e.g., `eyJ...`). 

This is particularly useful for developers who need to quickly grab JWT tokens, IDs, or specific field values without manually deleting the JSON structure (braces and keys) after pasting.

**Changes:**
- Added a `Row` to the `trailingBuilder` in `JsonPreviewer`.
- Included a new `IconButton` with the `Icons.abc` icon for copying the raw value.
- Updated the logic to use `node.value.toString()` for value extraction while maintaining the original `toJson(node)` logic for the full pair copy.
- Added descriptive tooltips for both buttons to improve accessibility.

## Related Issues

- Closes #638

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: This is a UI enhancement to the JSON Explorer trailing builder; existing logic for node extraction was utilized and manually verified with multiple data types.

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/widgets/previewer_json.dart
+++ b/lib/widgets/previewer_json.dart
@@ -272,28 +272,46 @@ class _JsonPreviewerState extends State<JsonPreviewer> {
                       trailingBuilder: (context, node) => node.isFocused
                           ? Padding(
                               padding: const EdgeInsets.only(right: 12),
-                              child: IconButton(
-                                padding: EdgeInsets.zero,
-                                constraints:
-                                    const BoxConstraints(maxHeight: 18),
-                                icon: const Icon(
-                                  Icons.copy,
-                                  size: 18,
-                                ),
-                                onPressed: () async {
-                                  final val = toJson(node);
-                                  String toCopy = '';
-                                  if (node.isClass ||
-                                      node.isArray ||
-                                      node.isRoot) {
-                                    toCopy = kJsonEncoder.convert(val);
-                                  } else {
-                                    toCopy = (val.values as Iterable)
-                                        .first
-                                        .toString();
-                                  }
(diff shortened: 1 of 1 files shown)


kpradyun:
https://github.com/user-attachments/assets/40254a95-46ec-4385-a489-3d1162d1270e

This is the change seen

PR #1087, opened 2026-02-20
fix: prevent duplicate history IDs in addHistoryRequest

Aman-Cool:
## PR Description

`addHistoryRequest` mutated `state` before checking `state == null`, making the null branch dead code. The else branch then spread `state!.keys` — which already contained the new ID — and appended it again, writing every history ID twice to Hive on every request sent. The UI looked correct because in-memory Maps deduplicate silently, but the on-disk `historyIds` list grew at 2× the correct rate permanently.

Fix: snapshot `existingKeys` from `state` before the mutation, build the persisted list from that snapshot, remove the dead ternary.

## Related Issues

NA

### Checklist

## Added/updated tests?

- [x] Yes

## OS on which you have developed and tested the feature

- [ ] Windows
- [x] macOS
- [ ] Linux

### DIFF
--- a/lib/providers/history_providers.dart
+++ b/lib/providers/history_providers.dart
@@ -78,13 +78,12 @@ class HistoryMetaStateNotifier
 
   void addHistoryRequest(HistoryRequestModel model) async {
     final id = model.historyId;
+    final existingKeys = state?.keys.toList() ?? [];
     state = {
       ...state ?? {},
       id: model.metaData,
     };
-    final List<String> updatedHistoryKeys =
-        state == null ? [id] : [...state!.keys, id];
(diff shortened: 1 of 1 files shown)


Aman-Cool:
Hi @animator ,Noticed that state was being mutated before the null check, which made the null branch unreachable; and since state!.keys already had the new id in it, the id was getting appended twice every time. snapshots existingKeys before the mutation so the list stays clean.

animator:
@Aman-Cool nice catch .. raise an issue with screenshots of debugPrint showing before and after the fix.
Link that issue to this PR.
Add some tests verifying the fix.

PR #1096, opened 2026-02-21
fix(ui): prevent empty text on Select Model button. Closes #942

### LINKED ISSUE #942: Empty text of Select model button when no model is chosen.
#### Describe the bug/problem
When selecting AI model, if user choses the model from the options of Ollama, OpenAI, Anthropic, Gemini, Azure OpenAI but not the exact model, the Select button displays empty text.

#### Expected behavior
It should continue to show Select Model text.

 - OS: Windows
 - Version: Windows 11 22H2

Image

saurabhhhcodes:
When selecting AI model, if user choses the model from the options of Ollama, OpenAI, Anthropic, Gemini, Azure OpenAI but not the exact model, the Select button previously displayed empty text. This PR explicitly checks for empty strings instead of just null values in the model name, falling back to 'Select Model'. Closes #942.

### DIFF
--- a/lib/screens/common_widgets/ai/ai_model_selector_button.dart
+++ b/lib/screens/common_widgets/ai/ai_model_selector_button.dart
@@ -45,7 +45,11 @@ class AIModelSelectorButton extends StatelessWidget {
               if (newAIRequestModel == null) return;
               onModelUpdated?.call(newAIRequestModel);
             },
-      child: Text(aiRequestModel?.model ?? 'Select Model'),
+      child: Text(
+        aiRequestModel?.model?.isNotEmpty == true
+            ? aiRequestModel!.model!
+            : 'Select Model',
+      ),
     );
   }
 }


animator:
Fix is already in pipeline.

saurabhhhcodes:
Hi @animator! This PR was originally opened explicitly to resolve the pre-existing **Issue #942**, which is already linked. I will add the requested UI screenshots showing the 'Select Model' button fix shortly!

saurabhhhcodes:
### Before (Bug)
The AI Model button shows blank/empty text when a provider is selected but no specific model is chosen.

### After (Fix)
The button now correctly displays "Select Model" as fallback text.

### Code Change
The root cause was that `model` could be an empty string `""` (not `null`), so the `??` operator didn't trigger the fallback. The fix adds an explicit `.isNotEmpty` check.

```diff
-      child: Text(aiRequestModel?.model ?? 'Select Model'),
+      child: Text(
+        aiRequestModel?.model?.isNotEmpty == true
+            ? aiRequestModel!.model!
+            : 'Select Model',
+      ),
```

PR #1130, opened 2026-02-23
feat: Auto-generate meaningful names for imported cURL and HAR requests

hihry:
## PR Description

What

When importing requests via cURL or HAR, the sidebar now shows short, readable names like GET /v1/users instead of "untitled" or the full raw URL.
## Related Issues

Why

Previously, importing a batch of cURL commands resulted in a sidebar full of "untitled" entries. HAR imports showed the entire URL including scheme, host, and query params — too long to scan. Users had to click each request just to figure out which was which.

https://github.com/user-attachments/assets/2917d86d-bf70-4e90-bf43-cbbfe65069f6

## Added/updated tests?

-  Yes
    
image

-  Windows

### DIFF
--- a/lib/importer/importer.dart
+++ b/lib/importer/importer.dart
@@ -1,4 +1,5 @@
 import 'package:apidash/consts.dart';
+import 'package:apidash/utils/http_utils.dart';
 import 'package:apidash_core/apidash_core.dart';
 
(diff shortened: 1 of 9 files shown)


hihry:
Oh my bad, a bit mistake

PR #1148, opened 2026-02-24
fix/remove-unused-iscompact-navbar

### LINKED ISSUE #1147: [FIX] Remove unused isCompact ternary logic from NavbarButton in button_navbar.dart
The `isCompact` parameter is declared and accepted in the `NavbarButton` constructor but never meaningfully used. It only appears in a redundant ternary where both branches return the same value, making it dead code.

`isCompact` parameter is accepted in the constructor but has no effect on the widget, making it misleading for developers who pass it expecting different behavior. This is a silent bug.

Image

**File:** `button_navbar.dart`

Replace `height: isCompact ? 36 : 36` with `height: 36`

hihry:
Fixes #1147

fixed the ternary logic  of isCompact

### DIFF
--- a/lib/screens/common_widgets/button_navbar.dart
+++ b/lib/screens/common_widgets/button_navbar.dart
@@ -49,7 +49,7 @@ class NavbarButton extends ConsumerWidget {
           mainAxisAlignment: MainAxisAlignment.center,
           children: [
             SizedBox(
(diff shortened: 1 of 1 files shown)


animator:
We are not making any code change wrt issue. There was some reason we had a ternary operator based on size. Requires more testing.

PR #1162, opened 2026-02-25
fix: escape regex special characters in env var substitution

### LINKED ISSUE #1164: Fix: escape regex special characters in environment variable keys
#### Describe the bug/problem
There was a logic bug in [lib/utils/envvar_utils.dart](cci:7://file:///d:/top/gsoc/New%20folder/test/apidash/lib/utils/envvar_utils.dart:0:0-0:0) where environment variable keys were being used to build a `RegExp` without being escaped. This caused two main issues:

1. **Substitution Failure:** If a variable name contains regex-special characters like `.` or [()]( the substitution fails. For example, a key named `auth(v2)` would prevent the value from being replaced because the regex engine interprets the parentheses as a capturing group instead of literal text.
2. **Security/Accuracy Risk:** A key like `api.url` was treated as a regex pattern where `.` matches any character. This could lead to accidental matching of incorrect strings like `api-url` or `api_url`.
3. **Stability:** Certain characters in a key name could potentially trigger a `FormatException` during regex construction, leading to unexpected app behavior.

#### Steps to Reproduce the bug/problem
1. Go to the **Variables** tab and create a new environment variable.
2. Set the key to a name with parentheses, e.g., `auth(v2)`, and give it a value.
3. Use this variable in a Request header or URL, e.g., `Bearer {{auth(v2)}}`.
4. Click **Send**.
5. Observe the response: The variable is not replaced and is sent as the literal string `{{auth(v2)}}`.

#### Expected behavior
Environment variable keys should always be treated as literal strings. By wrapping the key in `RegExp.escape()`, we ensure that characters like dots or parentheses are matched exactly as the user typed them.

#### Device Info
 - OS: Windows 11

#### Flutter Doctor
Please run the `flutter doctor -v` command and paste your output below:
## Screenshots
Image
Image

pratapsisodiya:
## PR Description

This PR fixes a critical bug in [substituteVariables] where environment variable keys were not being escaped before being used in a regular expression. 

**Key Fixes:**
- **Regex Character Escaping:** Utilized `RegExp.escape()` for each variable key. This ensures that variable names containing characters like `.` (dot), [()] (parentheses), or `[]` (brackets) are matched literally. 
- **Prevented Substitution Failures:** Previously, a variable named `auth(v2)` would fail to substitute because the regex engine interpreted the parentheses as a capturing group. After this fix, it matches exactly `{{auth(v2)}}`.
- **Improved Security/Stability:** Prevents unintentional matching (e.g., `api.url` matching `apiXurl`) and stops potential `FormatException` crashes caused by malformed key strings.

## Related Issues

- closes #1164 

### Checklist
- [x] I have updated my branch and synced it with project [main] branch before making this PR

## Added/updated tests?

- [x] Yes
- Added 3 new regression tests in [test/utils/envvar_utils_test.dart] to verify literal matching for dots, support for parentheses in keys, and ensuring no accidental pattern matches.

- [x] Windows
- [ ] macOS
- [ ] Linux

### DIFF
--- a/lib/utils/envvar_utils.dart
+++ b/lib/utils/envvar_utils.dart
@@ -44,7 +44,8 @@ String? substituteVariables(
   if (envVarMap.keys.isEmpty) {
     return input;
   }
-  final regex = RegExp("{{(${envVarMap.keys.join('|')})}}");
+  final escapedKeys = envVarMap.keys.map(RegExp.escape).join('|');
+  final regex = RegExp("{{($escapedKeys)}}");
 
   String result = input.replaceAllMapped(regex, (match) {
     final key = match.group(1)?.trim() ?? '';
--- a/test/utils/envvar_utils_test.dart
+++ b/test/utils/envvar_utils_test.dart
@@ -182,6 +182,27 @@ void main() {
           expected);
     });
 
+    test('substituteVariables with dot in variable name (api.url)', () {
+      const input = 'https://{{api.url}}/endpoint';
+      final envMap = {'api.url': 'example.com'};
+      expect(
+          substituteVariables(input, envMap), 'https://example.com/endpoint');
+    });
+
+    test('substituteVariables does NOT substitute similar‑looking keys', () {
+      const input = 'https://{{apiXurl}}/endpoint';
+      final envMap = {'api.url': 'example.com'};
+      expect(
+          substituteVariables(input, envMap), 'https://{{apiXurl}}/endpoint');
+    });
+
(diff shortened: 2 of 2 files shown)


pratapsisodiya:
## Before

https://github.com/user-attachments/assets/8c22c099-e9b3-427c-a0a1-b488f8e0e5b5

## After

https://github.com/user-attachments/assets/295ca84e-ed50-4fa5-b5d4-5fe7f9cb228c

PR #1245, opened 2026-03-02
feat: auto-focus URL field when creating a new request

### LINKED ISSUE #1244: Auto-focus URL field when creating a new request

When creating a new request using the `+` button, the URL input field is not automatically focused. This requires an additional mouse click before typing the URL.

For users who frequently create new requests, this slightly interrupts the workflow and makes the interaction less keyboard-friendly.

**Screen recording demonstrating current behavior:**
https://github.com/user-attachments/assets/fdd96848-1976-4e0c-b8cc-b627cfd614df

After a new request is created:

* The URL input field should automatically receive focus.
* The cursor should be placed inside the URL field so the user can immediately start typing.
* This should only trigger when a new request is added (not when switching between existing requests).

This would improve usability and make request creation faster and more intuitive.

This small improvement would make the request creation flow smoother and more aligned with common developer tools, where the primary input field is focused automatically after creating a new item.

Aanish-Bangre:
## PR Description

This PR improves the request creation workflow by automatically focusing the URL input field when a new request is added.

Currently, after pressing the `+` button, users must manually click the URL field before typing. With this change:

* The URL field receives focus immediately after a new request is created.
* The cursor is placed inside the field so users can start typing instantly.
* Focus is only triggered when adding a new request (not when switching between existing ones).

This makes the workflow smoother and more keyboard-friendly.

### Before

[https://github.com/user-attachments/assets/ae2d749c-a966-494e-ad3e-e0ec882166fc](https://github.com/user-attachments/assets/ae2d749c-a966-494e-ad3e-e0ec882166fc)

### After

[https://github.com/user-attachments/assets/6824eb10-5e70-447d-aa4b-2f8b365661c2](https://github.com/user-attachments/assets/6824eb10-5e70-447d-aa4b-2f8b365661c2)

## Related Issues

Closes #1244

### Checklist

* [x] I have gone through the contributing guide
* [x] I have updated my branch and synced it with project `main` branch before making this PR
* [x] I am using the latest Flutter stable branch (run `flutter upgrade` and verify)
* [x] I have run the tests (`flutter test`) and all tests are passing

## Added/updated tests?

* [x] No, and this is why: This change affects UI focus behavior only and does not modify business logic. Existing tests pass without modification.

* [x] Windows
* [ ] macOS
* [ ] Linux

### DIFF
--- a/lib/providers/ui_providers.dart
+++ b/lib/providers/ui_providers.dart
@@ -33,12 +33,6 @@ final nameTextFieldFocusNodeProvider =
   return focusNode;
 });
 
-final urlFieldFocusNodeProvider = Provider.autoDispose<FocusNode>((ref) {
-  final focusNode = FocusNode();
-  ref.onDispose(focusNode.dispose);
-  return focusNode;
-});
-
 final collectionSearchQueryProvider = StateProvider<String>((ref) => '');
 final environmentSearchQueryProvider = StateProvider<String>((ref) => '');
 final importFormatStateProvider =
--- a/lib/screens/common_widgets/env_trigger_field.dart
+++ b/lib/screens/common_widgets/env_trigger_field.dart
@@ -59,8 +59,12 @@ class EnvironmentTriggerFieldState extends State<EnvironmentTriggerField> {
 
   @override
   void dispose() {
-    controller.dispose();
-    _focusNode.dispose();
+    if (widget.controller == null) {
+      controller.dispose();
+    }
+    if (widget.focusNode == null) {
+      _focusNode.dispose();
+    }
     super.dispose();
   }
 
--- a/lib/screens/home_page/collection_pane.dart
+++ b/lib/screens/home_page/collection_pane.dart
@@ -28,9 +28,6 @@ class CollectionPane extends ConsumerWidget {
           SidebarHeader(
             onAddNew: () {
               ref.read(collectionStateNotifierProvider.notifier).add();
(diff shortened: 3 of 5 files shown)


animator:
What happens in mobile/smaller widths as request addition happens in the drawer?

Aanish-Bangre:
Good point. I tested the behavior on smaller widths where request creation happens through the drawer.

I updated the implementation so focus handling is reactive instead of using the earlier signal counter approach.

• The URL field now listens to selectedIdStateProvider and focuses when a request becomes active.
• On mobile, focus is skipped while the drawer is open.
• When the drawer closes (leftDrawerStateProvider), the URL field focuses if the current request has an empty URL.

I also fixed a dispose issue in env_trigger_field.dart where externally provided FocusNodes were always being disposed, which could cause crashes during rebuilds.

demo video - 

[https://github.com/user-attachments/assets/2e3a4107-871d-479b-8704-89cec305143c](https://github.com/user-attachments/assets/2e3a4107-871d-479b-8704-89cec305143c)

animator:
You have broken the version control history by rewriting the PR. 
I wanted to review the changes before and after my comment.

Aanish-Bangre:
@animator 
Sorry about that — I rebased and force-pushed the branch while updating the PR and didn’t realize it would remove the previous commit history.

Would you prefer that I open a new PR with the fixes applied as additional commits?

PR #1290, opened 2026-03-06
feat: add custom OpenAI-compatible LLM provider support

### LINKED ISSUE #1175: Support for Custom OpenAI-Compatible LLM Providers (Groq, OpenRouter, Mistral, etc.)
Currently, APIDash only supports specific built-in AI providers, including OpenAI, Anthropic, Gemini, Azure OpenAI, and Ollama, for its GenAI features. I want to use other alternative LLM providers that offer OpenAI-compatible API endpoints, such as OpenRouter, Groq, Mistral, Together AI, and LlamaAPI. Since there is no option to specify a custom endpoint and model identifier, I cannot connect APIDash's AI features to these powerful, often cost-effective or faster alternative models.

I would like a **"Custom (OpenAI Compatible)"** option added to the AI Model Selector Dialog. 

When this provider is selected, the dialog should allow the user to provide:  
1. **API Key / Credential:** The API key for the custom provider.  
2. **Endpoint:** The full OpenAI-compatible chat completions URL, for example, `https://api.groq.com/openai/v1/chat/completions` `https://openrouter.ai/api/v1/chat/completions` , `https://api.mistral.ai/v1/chat/completions`.  
3. **Model Name:** A text field to type out the identifier of the specific model to be used, for example, `llama-3.3-70b-versatile`.  

This would use the existing [OpenAIModel]implementation structure under the hood to serialize the request correctly using Bearer authentication and the standard OpenAI JSON schema, but it would point to the custom URL provided by the user.

This feature would greatly increase the accessibility of AI integrations in APIDash by allowing access to hundreds of open-source and alternative hosted models without needing to change the code for every new provider on the market.

veyron-kairo:
## Summary
- Adds a new `Custom OpenAI-Compatible` provider so users can connect to any OpenAI-compatible API like Groq, OpenRouter, Mistral, etc.
- Users can enter their own endpoint URL, API key, and model identifier instead of picking from a fixed list
- Follows the same request/response format as the existing OpenAI provider with Bearer token auth

## Changes
- New `CustomOpenAIModel` provider class in `packages/genai`
- Added `customOpenai` enum value to `ModelAPIProvider` and wired it up in `kModelProvidersMap`
- Updated `models.json` and regenerated models data
- Updated the AI Model Selector dialog to show a text input for model ID when using the custom provider

Closes #1175


veyron-kairo:
Hey @ashitaprasad @animator, this implements #1175 — adds a Custom OpenAI-Compatible provider so users can connect to any OpenAI-compatible API (Groq, OpenRouter, Mistral, etc). Happy to make changes if needed!

Issue #1296, opened 2026-03-07
Bug: Dashbot Ollama model Integration: "400 Invalid Request" & Configuration Wiped on Selection


armanraymagit:
# [Bug Report] Dashbot Ollama Integration: "400 Invalid Request" & Configuration Wiped on Selection

#### Bug/Problem
When using Dashbot and attempting to select or configure an Ollama model (e.g., `qwen 2.5 `), I am encountering a `400 Invalid Request: model is required` API error. 

Additionally, anytime I try to open the selector and pick an Ollama model, my existing custom configurations—specifically the **Custom URL** and **API Key**—are actively wiped out and reset within the UI. This prevents any successful interaction with custom local models since their endpoint parameters disappear before execution.

**Component:** Dashbot / AI Model Selector Dialog (`lib/screens/common_widgets/ai/ai_model_selector_dialog.dart`)

**Suspected Root Cause:**
The core issue appears to stem from how the `AIModelSelectorDialog` handles state changes when a user selects a new AI provider or model from the dropdown.

When the `onChanged` event fires, the logic instantiates a completely **new** `AIRequestModel` object from scratch:

```dart
AIRequestModel newModel = AIRequestModel(
  aiModelProvider: value.provider,
  model: aiModel.name,
);
```

Because it creates a brand new instance, it only evaluates the `aiModelProvider` and the new `model` name. It appears to completely ignore the existing state of the dialog. Consequently, user-defined properties like the `url` and `apiKey` are abandoned and implicitly set to `null` on the new object. 

When this truncated model is eventually passed back down to Dashbot to execute via the `ollama_dart` client, the lack of a proper endpoint URL triggers the `400` errors from the server.

**Console Logs:**
```text
[ChatBubble] Actions count: 0 | msg: hey
[ChatBubble] Actions count: 0 | msg: No response received from the AI.
[ChatBubble] Actions count: 0 | msg: hey
[ChatBubble] Actions count: 0 | msg: No response received from the AI.
LLM_EXCEPTION: 400
{"error":{"message":"model is required","type":"api_error","param":null,"code":null}}
```

**Local Ollama Server Logs:**
```text
[GIN] 2026/03/07 - 08:31:30 | 200 |    2.8718ms | 127.0.0.1 | GET  "/api/ps"
[GIN] 2026/03/07 - 08:32:05 | 400 |   44.3568ms | 127.0.0.1 | POST "/v1/chat/completions"
```

#### Steps to Reproduce the bug/problem
[...]

PR #1304, opened 2026-03-07
Fix Dashbot AI integration bug with Ollama provider #1296

VanshKaushal:
Fix Dashbot AI Model Selector state reset and improve Ollama compatibility #1296 
Problem

Selecting a new AI model in the AIModelSelectorDialog caused previously entered configuration values (such as Custom URL and API Key) to be cleared. This resulted in malformed requests to local providers like Ollama, leading to errors such as:

400 {"error":{"message":"model is required"}}
Root Cause

The dialog recreated a new AIRequestModel instance on every model change instead of updating the existing state. This discarded previously entered configuration fields.

Fix

- Introduced copyWith() updates when switching models to preserve existing configuration

- Removed ValueKey usage in BoundedTextField which triggered widget rebuilds that reset text controllers
 
- Added strict validation in ai_request_utils.dart to ensure model and url are present before executing requests
 
- Improved Ollama request handling by dynamically formatting payloads for /api/chat and /api/generate
 
- Enhanced error propagation to surface server-side API errors directly in the UI

Result

- Custom configuration is preserved when switching AI models

- Ollama integrations work correctly across supported endpoints
 
- Users now receive meaningful error messages instead of generic failures

### DIFF
--- a/lib/screens/common_widgets/ai/ai_model_selector_dialog.dart
+++ b/lib/screens/common_widgets/ai/ai_model_selector_dialog.dart
@@ -68,8 +68,16 @@ class _AIModelSelectorDialogState extends ConsumerState<AIModelSelectorDialog> {
                           onChanged: (x) {
                             setState(() {
                               selectedProvider = x;
-                              newAIRequestModel = mappedData[selectedProvider]
+                              final defaultModel = mappedData[selectedProvider]
                                   ?.toAiRequestModel();
+                              if (newAIRequestModel != null) {
+                                newAIRequestModel = newAIRequestModel!.copyWith(
+                                  modelApiProvider: selectedProvider,
+                                  model: defaultModel?.model,
+                                );
+                              } else {
+                                newAIRequestModel = defaultModel;
+                              }
                             });
                           },
                           value: selectedProvider,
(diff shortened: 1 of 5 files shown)


VanshKaushal:
Hi

I’ve opened this PR (#1304) to address the issue where selecting a different AI model in the `AIModelSelectorDialog` would reset existing configuration values like the custom URL and API key. While reproducing the bug, I noticed that the dropdown handler was creating a new `AIRequestModel` instance each time a model was selected, which caused the previously entered configuration to be lost. This also resulted in malformed requests when using local providers like Ollama (e.g., `400: model is required`).

In this PR, I updated the logic to use `AIRequestModel.copyWith()` so that the existing configuration is preserved when switching models. I also removed a `ValueKey` on `BoundedTextField` that was triggering widget rebuilds and clearing active text inputs.

Along with the main fix, I added validation to ensure `model` and `url` are present before sending requests, improved request formatting for Ollama endpoints (`/api/chat` and `/api/generate`), and surfaced API error messages more clearly in the UI.

All tests are passing locally. Please let me know if there are any changes or improvements you’d like me to make. Happy to update the PR as needed. Thanks for reviewing!

Issue #1330, opened 2026-03-11
[FEATURE] : Enhance Postman Collection Importer


hihry:
## Feature Request: Enhance Postman Collection Importer — Auth, Variables & `urlencoded` Body

The Postman v2.1 importer silently drops three key data categories:

**1. Authentication** — The `auth` block (Bearer, Basic, API Key, OAuth, Digest) is fully ignored.
The `Request` model has no `auth` field and no mapping to API Dash's existing `AuthModel` types exists.

**2. `urlencoded` Body Mode** — Only `raw` and `formdata` are handled. The `urlencoded` mode —
standard for OAuth token endpoints and form submissions — is silently dropped, leaving requests with no body.

**3. Collection Variables** — `{{variable}}` placeholders are never resolved. `PostmanCollection`
has no `variable` field, so URLs, headers, and bodies retain broken placeholders like `{{base_url}}`.

### Proposed Fixes

- **Auth:** Add `Auth` + `AuthKeyValue` freezed models; add `Auth? auth` to `Request`; map Postman auth types to existing `AuthBearerModel`, `AuthBasicAuthModel`, `AuthApiKeyModel`, etc. in `postman_io.dart`
- **`urlencoded`:** Add `List<Urlencoded>? urlencoded` to `Body`; map items to `FormDataModel` entries with `FormDataType.text` and `ContentType.formdata`
- **Variables:** Add `List<Variable>? variable` to `PostmanCollection`; build a `Map<String, String>` and substitute `{{key}}` placeholders in url, headers, and body before conversion

### Affected Files

| File | Change |
|---|---|
| `postman_collection.dart` | Add `Auth`, `AuthKeyValue`, `Urlencoded`, `Variable` models; extend `Request`, `Body`, `PostmanCollection` |
| `postman_collection.freezed.dart` / `.g.dart` | Regenerate via `build_runner` |
| `postman_io.dart` | Add auth mapping, `urlencoded` handling, variable substitution |
| `postman_test.dart` | Add tests for all three new scenarios |

> **Note:** The importer currently has only 4 serialization tests and zero conversion-logic coverage.
> The Insomnia importer has the same auth gap and could be addressed in the same effort.

VanshKaushal:
hello @animator , I have a pretty good idea and proposal to solve this issue , I can contribute a bullseye PR to fix it .

Roopalgn:
Hi! I’ve been exploring the Postman importer implementation and looking into this issue.

[...]

PR #1356, opened 2026-03-17
extend Postman collection models with auth, variable, urlencoded, and…

hihry:
Fixes : #1330 

This PR fixes major Postman import gaps by preserving authentication, urlencoded body fields, and collection variable substitution during import.
I verified all test cases are passing.


hihry:
Hi @animator , I noticed the `gsoc26-idea` label was added to PR #1356. Since this was an independently identified issue and not part of the official GSoC idea list, I wanted to check, whether is it being tracked separately as a potential GSoC scope extension? Would love some clarity so I can plan accordingly. Thanks!

PR #1384, opened 2026-03-20
feat: detect MIME type from magic bytes for application/octet-stream

### LINKED ISSUE #1382: Previewer fallback based on magic bytes

hihry:
Fixes #1382 
<html>
<body>
<html><head></head><body><h1>feat: detect MIME type from magic bytes for <code>application/octet-stream</code> responses</h1>
<h2>Problem</h2>
<p>When a server responds with <code>Content-Type: application/octet-stream</code>, API Dash cannot determine the actual file format, so the Preview tab either fails or shows a fallback error instead of rendering the file correctly.</p>
<h2>Solution</h2>
<p>Sniff the first few bytes of the response body (magic bytes) to detect the true MIME type and route it to the correct previewer.</p>
<h2>Changes</h2>

File | What
-- | --
lib/utils/magic_bytes_utils.dart | New utility — detects MIME from magic bytes (PNG, JPEG, GIF, BMP, WebP, PDF, MP3, WAV, MP4)
lib/widgets/previewer.dart | Uses resolved MIME type for preview routing instead of raw header
lib/widgets/response_body.dart | Derives effective MediaType from sniffing so correct UI tabs appear
test/utils/magic_bytes_utils_test.dart | Unit tests for all supported formats
test/widgets/previewer_test.dart | Sniffing test — octet-stream + JPEG bytes → image preview
test/widgets/response_widgets_test.dart | UI test — Preview tab appears for recognizable binary content

<h2>Testing</h2>
<p>Send a GET request to any URL that returns a binary file with <code>Content-Type: application/octet-stream</code> and confirm the Preview tab renders correctly.</p>
<pre><code>https://raw.githubusercontent.com/mozilla/pdf.js/master/web/compressed.tracemonkey-pldi-09.pdf
</code></pre></body></html>
</body>

https://github.com/user-attachments/assets/52582eaa-7db9-4855-b31a-af57ae029d38
<p>In this sample test I just send a request and imported a pdf of Content-Type: application/octet-stream, previously users were encountring a error message. </p>
</html>

### DIFF
--- a/lib/utils/magic_bytes_utils.dart
+++ b/lib/utils/magic_bytes_utils.dart
@@ -1,0 +1,48 @@
+import 'dart:typed_data';
+
+(String, String) detectMimeTypeFromBytes(Uint8List bytes) {
+  if (bytes.length < 4) return ('application', 'octet-stream');
+
+  // PNG: 89 50 4E 47
+  if (bytes[0] == 0x89 && bytes[1] == 0x50 &&
+      bytes[2] == 0x4E && bytes[3] == 0x47) return ('image', 'png');
+
+  // JPEG: FF D8 FF
+  if (bytes[0] == 0xFF && bytes[1] == 0xD8 && bytes[2] == 0xFF)
+    return ('image', 'jpeg');
+
+  // GIF: 47 49 46 38
+  if (bytes[0] == 0x47 && bytes[1] == 0x49 &&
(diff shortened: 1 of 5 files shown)


animator:
Too much code changes. Solution can be simpler.

hihry:
Thanks for feedback, I will work upon it and provide a better optimize solution for it.

PR #1393, opened 2026-03-21
Route URL-field cURL paste through DashBot import

### LINKED ISSUE #953: Auto-parse curl command when pasted into URL field

When working with APIs, I often have curl commands available (from backend teams, docs, or logs).
Currently, when I paste a curl command into the URL field in API Dash, it is treated as plain text,
and I have to manually extract the URL, headers, method, and body and fill them into the request editor.

This makes importing existing curl commands time-consuming and error-prone compared to tools like Postman.

Add automatic curl command parsing when a user pastes a curl command into the URL field.

When a curl command is detected on paste:
- Parse the curl using the existing CurlIO parser
- Automatically populate request method, URL, headers, query parameters, body, and form data
- Replace the URL text field content with only the parsed URL (not the full curl command)
- Show a user-friendly error message if the curl command is invalid

This behavior should be limited to REST API requests and should not trigger during normal typing.

This feature would significantly improve developer experience and bring API Dash closer to the
workflow users expect from modern API clients like Postman, especially for quickly importing
existing requests.

Adi-gitX:
Closes #953

## Summary
This PR routes pasted cURL commands from the REST URL field into the existing DashBot import-cURL workflow.

Instead of leaving a pasted `curl ...` command in the URL field, API Dash now opens DashBot, pins the chat route, starts the import-cURL task, and forwards the pasted command through the current DashBot parsing path.

## Root Cause
The editor URL field accepted pasted cURL commands as plain text, but the actual parsing and request-conversion flow already exists in DashBot. That left users with a broken editor value and duplicated the work needed to import a request correctly.

## Fix
I added a paste-aware interception hook to the URL field path and used it only for the REST request editor.

The interception is clipboard-checked so normal typing is not rerouted. When the pasted text starts with `curl `, the field now:
- opens DashBot if needed
- pins DashBot to chat
- triggers the existing import-cURL task prompt
- forwards the pasted cURL text into the current DashBot import flow
- restores the URL field to its previous value instead of leaving raw cURL there

## Validation
Passed locally:
- `flutter test test/screens/common_widgets/env_trigger_field_test.dart`
- `flutter test test/widgets/textfields_test.dart`
- `flutter test test/dashbot/features/chat/viewmodel/chat_viewmodel_test.dart`

## AI Usage
AI assistance was used for exploration and drafting, and the final code was reviewed manually before submission.

### DIFF
--- a/lib/screens/common_widgets/env_trigger_field.dart
+++ b/lib/screens/common_widgets/env_trigger_field.dart
@@ -1,4 +1,6 @@
+import 'dart:async';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:multi_trigger_autocomplete_plus/multi_trigger_autocomplete_plus.dart';
 import 'package:extended_text_field/extended_text_field.dart';
 import 'env_regexp_span_builder.dart';
@@ -13,6 +15,7 @@ class EnvironmentTriggerField extends StatefulWidget {
     this.focusNode,
     this.onChanged,
     this.onFieldSubmitted,
+    this.onPastedText,
     this.style,
     this.decoration,
     this.optionsWidthFactor,
@@ -30,6 +33,7 @@ class EnvironmentTriggerField extends StatefulWidget {
   final FocusNode? focusNode;
   final void Function(String)? onChanged;
   final void Function(String)? onFieldSubmitted;
+  final Future<bool> Function(String)? onPastedText;
   final TextStyle? style;
   final InputDecoration? decoration;
   final double? optionsWidthFactor;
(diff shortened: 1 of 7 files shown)


Adi-gitX:
Addressed the valid Copilot review points in the latest push: nullable `initialValue` is handled safely on request switches, internally owned controllers are disposed before replacement, handled cURL pastes no longer emit transient raw `curl ...` values through `onChanged`, and the async paste path now guards against exceptions and stale restores. I did not remove the `apidash_design_system` import because `kCodeStyle` is still used from that package in `field_url.dart`.

Adi-gitX:
Added a Visual Proof section with animated before/after assets to make the cURL-paste flow easier to review.

Adi-gitX:
Removed the earlier placeholder visual assets. They were not captured from a live API Dash runtime, and I do not want synthetic proofs in the PR. A real demo from the running app requires a full Xcode-based macOS setup on this machine.

PR #1398, opened 2026-03-21
Allow choosing response download location

### LINKED ISSUE #1395: Allow users to choose save location before downloading response
I want to download API responses and save them in a specific folder for better organization.

Currently, when I click download, the file is automatically saved to a default location without giving me an option to choose where to save it. This makes it difficult to manage files, especially when downloading multiple responses.

It would be helpful to provide a file picker dialog before downloading the response, allowing users to choose the save location and optionally rename the file.

This would improve usability and align with standard desktop application behavior.

This issue was observed while using the Windows desktop version of the application.

I’d be happy to work on this if it’s open.

Adi-gitX:
Closes #1395

Summary
This PR lets users choose a save location before downloading a response body from API Dash.

Root Cause
Response downloads always wrote directly to the default downloads directory, so users had no chance to choose a folder or rename the file during the save flow.

Fix
- added a save-location helper backed by `file_selector`
- kept the existing downloads-directory behavior as the default path
- enabled the save dialog only for response downloads so the change stays narrow
- added utility and widget coverage for the new response download flow

Validation
Passed locally:
- `flutter test test/utils/file_utils_test.dart`
- `flutter test test/widgets/button_save_download_test.dart`
- `flutter test test/widgets/response_widgets_test.dart`

AI Usage
AI assistance was used for exploration and drafting, and the final code and tests were reviewed manually before submission.

### DIFF
--- a/lib/utils/save_utils.dart
+++ b/lib/utils/save_utils.dart
@@ -37,6 +37,7 @@ Future<void> saveToDownloads(
       ? await getFileSavePath(name, resolvedExt, mimeType: mimeType)
       : await getFileDownloadpath(name, resolvedExt);
   if (chooseSaveLocation && path == null) {
+    sm.hideCurrentSnackBar();
     return;
   }
   if (path != null) {
--- a/test/widgets/button_save_download_test.dart
+++ b/test/widgets/button_save_download_test.dart
@@ -56,6 +56,21 @@ class FakeFileSelectorPlatform extends FileSelectorPlatform {
   }
 }
 
+Future<bool> waitForFileToExist(
+  String path, {
+  Duration timeout = const Duration(seconds: 1),
+  Duration pollInterval = const Duration(milliseconds: 10),
+}) async {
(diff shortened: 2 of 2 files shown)


Adi-gitX:
Addressed the follow-up review points in the latest push: canceling the save dialog now clears any stale snackbar, and the widget test no longer relies on a fixed sleep. Revalidated with .

Adi-gitX:
Addressed the follow-up review points in the latest push: canceling the save dialog now clears any stale snackbar, and the widget test no longer relies on a fixed sleep. Revalidated with the response download widget and utility tests.

Adi-gitX:
Added a Visual Proof section with animated before/after assets so the save-location flow is visible in the PR.

Adi-gitX:
Removed the earlier placeholder visual assets. They were not captured from a live API Dash runtime, and I do not want synthetic proofs in the PR. A real demo from the running app requires a full Xcode-based macOS setup on this machine.

PR #1413, opened 2026-03-22
fix: detect magic bytes for application/octet-stream previewer #1382

### LINKED ISSUE #1382: Previewer fallback based on magic bytes

gauravk16in:
## PR Description

When a file is served with the generic application/octet-stream MIME type, it indicates an unknown binary file. Magic bytes - a unique sequence of bytes at the beginning of the file to identify its true format and select an appropriate previewer.
_Now any `application/octet-stream` MIME type can be previewed by identifying the first sequence of the binary file type._

**Video before the feature-**

https://github.com/user-attachments/assets/26a43aee-17a1-4e1a-998d-f5b46129ed07

**Video after adding the feature-**

https://github.com/user-attachments/assets/2ecce19e-2452-42fb-80a7-d77d506a45f8

@animator Please look into this pr and merge it.
thank you!

## Related Issues

- Closes #1382

### Checklist
- [X] I have updated my branch and synced it with project `main` branch before making this PR
- [X] I am using the latest Flutter stable branch (run `flutter upgrade` and verify)

## Added/updated tests?

- [X] Yes

- [ ] Windows
- [ ] macOS
- [X] Linux

### DIFF
--- a/lib/widgets/previewer.dart
+++ b/lib/widgets/previewer.dart
@@ -138,11 +138,91 @@ class _PreviewerState extends State<Previewer> {
         );
       }
     }
-    var errorText = errorTemplate.render({
-      "showRaw": widget.hasRaw,
-      "showContentType": true,
-      "type": "${widget.type}/${widget.subtype}",
-    });
-    return ErrorMessage(message: errorText);
+    // var errorText = errorTemplate.render({
+    //   "showRaw": widget.hasRaw,
+    //   "showContentType": true,
+    //   "type": "${widget.type}/${widget.subtype}",
+    // });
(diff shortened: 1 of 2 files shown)


animator:
There is an easier path forward with less code changes.

gauravk16in:
@animator definitely I will try one more time or may be two until I get a better approach and reduce the line of codes by using any existing helper function and importing it directly to the `previewer.dart` btw this approach also taught me something interesting that I can do with ASCII values and ya this was the one of the great issue and I was curious to solve this from the day 1 when you posted the issue.
Thanks for your suggestion.

PR #1414, opened 2026-03-22
Improve API error messages for better user understanding

### LINKED ISSUE #1412: Improve error messages for better user understanding
While installing and exploring API Dash locally, I noticed that some API error messages are not easy to understand at first glance.

For example, errors like Failed host lookup or request issues show technical messages that require the user to search and understand what the problem actually is.

It would be helpful if the app could show simpler and more user-friendly messages along with the technical error. This would make it easier for users to quickly understand what went wrong without confusion and also would improve usability, especially for new users.

sakshi2255:
## Overview
Improved error messages shown to users when API requests fail.

## Problem
Previously, raw technical errors like "SocketException" were displayed, which are not easy for users to understand.

## Solution
Added user-friendly error messages for common failure scenarios such as:
- Network issues
- Invalid API URLs
- Connection failures

Now users see clear messages like:
"Unable to connect to server. Please check your internet or API URL."

## Testing
Tested with:
- Invalid API URL
- No internet connection
- Valid API request

All scenarios now display meaningful messages.

### Before
before

### After
after

Fixes #1412

### DIFF
--- a/lib/providers/collection_providers.dart
+++ b/lib/providers/collection_providers.dart
@@ -22,6 +22,7 @@ final selectedRequestModelProvider = StateProvider<RequestModel?>((ref) {
   }
 });
 
+
 final selectedSubstitutedHttpRequestModelProvider =
     StateProvider<HttpRequestModel?>((ref) {
       final selectedRequestModel = ref.watch(selectedRequestModelProvider);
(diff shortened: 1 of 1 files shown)


sakshi2255:
Tested with invalid URL and no internet scenarios. Now showing user-friendly messages instead of technical errors.

PR #1429, opened 2026-03-23
Add auto-focus to URL field when creating a new request

tech-nitin:
Fixes issue where the URL input field is not automatically focused when creating a new request.

Added autofocus to the TextFormField so users can immediately start typing without extra clicks.

This improves usability and makes the workflow more efficient.


tech-nitin:
Hi, I’ve fixed this issue. Please review my PR.

PR #1443, opened 2026-03-23
fix(better_networking): handle query-only input in stripUrlParams

yassinekolsi:
## Summary
Fixes an edge case in URL utility behavior where stripUrlParams did not strip query strings when the query delimiter was at index 0.

## Root cause
stripUrlParams only stripped when indexOf('?') was greater than 0. For query-only inputs like ?x=1, the index is 0, so the query was incorrectly preserved.

## Changes
- Updated stripUrlParams in better_networking URI utils:
  - from idx > 0 to idx >= 0
- Added tests in uri_utils_test:
  - query-only input returns empty string
  - fragment before query delimiter is preserved correctly

## Files changed
- packages/better_networking/lib/utils/uri_utils.dart
- packages/better_networking/test/utils/uri_utils_test.dart

## Local validation
Commands run locally:
- flutter test packages/better_networking/test/utils/uri_utils_test.dart
- flutter test packages/better_networking
- flutter analyze packages/better_networking

Results:
- Both test commands passed.
- Analyze reports 3 pre-existing info-level lints unrelated to this change.

## Risk
Low risk, one-line logic fix plus regression coverage. No API changes.

### DIFF
--- a/packages/better_networking/lib/utils/uri_utils.dart
+++ b/packages/better_networking/lib/utils/uri_utils.dart
@@ -17,6 +17,11 @@ String stripUriParams(Uri uri) {
   return "${uri.scheme}://${uri.authority}${uri.path}";
 }
 
+/// Removes everything from the first `?` onwards using string splitting.
+///
+/// This helper is intentionally string-based and does not enforce strict URI
+/// semantics. For example, a `?` that appears after `#` is still treated as a
+/// split point.
 String stripUrlParams(String url) {
   var idx = url.indexOf("?");
   return idx >= 0 ? url.substring(0, idx) : url;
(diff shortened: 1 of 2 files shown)


yassinekolsi:
Thanks for the sharp catch on URI semantics.

You are right that in strict RFC terms, a  after  is fragment content. For this helper, the intended contract is lightweight string splitting on the first  rather than full URI parsing.

I pushed a follow-up commit to make that explicit:
- Added a doc comment on stripUrlParams clarifying string-based behavior.
- Updated the related test wording to match that contract.

Validation rerun locally after the update:
- flutter test packages/better_networking/test/utils/uri_utils_test.dart passed
- flutter test packages/better_networking passed

yassinekolsi:
Small correction to my previous comment for readability:

In strict RFC terms, a question-mark character appearing after a hash character is fragment content. For this helper, the intended contract is lightweight string splitting at the first question-mark character, not full URI parsing.

The follow-up commit already makes this explicit in code docs and test wording.

PR #1446, opened 2026-03-23
fix(autocomplete): guard trigger scan against out-of-range cursor

yassinekolsi:
## Summary
Prevents out-of-range cursor crashes in trigger detection for `multi_trigger_autocomplete_plus`.

## Root cause
`invokingTrigger` used `text.substring(0, selection.baseOffset)` directly. If the cursor offset is greater than text length, this throws `RangeError`.

## Changes
- Clamp cursor position to valid bounds before substring operations.
- Added regression test with out-of-bounds cursor offset.

## Files changed
- `packages/multi_trigger_autocomplete_plus/lib/src/autocomplete_trigger.dart`
- `packages/multi_trigger_autocomplete_plus/test/autocomplete_trigger_test.dart`

## Local validation
- `flutter test packages/multi_trigger_autocomplete_plus/test/autocomplete_trigger_test.dart` (passed)
- `flutter test packages/multi_trigger_autocomplete_plus` (passed)
- `flutter analyze packages/multi_trigger_autocomplete_plus` (existing info-level lints only)

## Risk
Low risk; behavior is unchanged for valid offsets and now safe for malformed/out-of-range offsets.

### DIFF
--- a/packages/multi_trigger_autocomplete_plus/test/autocomplete_trigger_test.dart
+++ b/packages/multi_trigger_autocomplete_plus/test/autocomplete_trigger_test.dart
@@ -105,11 +105,11 @@ void main() {
           selection: TextSelection.collapsed(offset: 100),
         );
 
-        expect(() => trigger.invokingTrigger(value), returnsNormally);
         final invoked = trigger.invokingTrigger(value);
 
         expect(invoked, isNotNull);
         expect(invoked!.query, 'Sahil');
+        expect(invoked.selection.extentOffset, text.length);
       },
     );
   });


yassinekolsi:
Implemented the Copilot suggestion in the out-of-bounds cursor test.

What changed:
- Removed the redundant first invocation and now call invokingTrigger exactly once.
- Added an explicit assertion that the returned selection is clamped: invoked.selection.extentOffset == text.length.

Validation run locally:
- flutter test test/autocomplete_trigger_test.dart passed
- flutter test (package) passed
- flutter analyze reports only pre-existing info-level lints (no new issues introduced by this change)

Issue #1269, opened 2026-03-05
Revamp Model Selector dialog


animator:
Meta list to track issues with model selector dialog.
Currently, it needs a complete revamp.

codealchemist007:
Hi @animator,

I’m a Flutter developer and would like to take on the sub-issue: Support for Custom OpenAI-Compatible LLM Providers (#1175). I’ve recently been working with LLM integrations and have a good understanding of OpenAI-compatible API structures.

My initial thoughts on the approach:

* UI: I'll extend the Model Selector dialog to include the "Custom (OpenAI Compatible)" option.
* State Management: I plan to add fields for the custom Endpoint and Model Name, ensuring they are persisted correctly.
* Logic: Since the goal is to leverage the existing OpenAIModel structure, I'll look into abstracting the base URL so it can be overridden by the user-provided endpoint while maintaining the standard Bearer authentication and JSON schema.

I’m ready to dive into the codebase and start on a draft PR. Could you please assign this issue to me?

Thanks!

saurabh24thakur:
Hi @animator, I've been keeping an eye on the Model Selector revamp! I've worked quite a bit with Groq and other LLM integrations in my own projects—like an AI-driven interview platform—so I’d love to help out with this.

I see that #1175  is already being discussed, so I’m happy to pick up one of the other sub-tasks to keep things moving. I could take on #1183  to add validation for API keys and endpoints, or dive into the Ollama bug (#1296 ) to make sure the configuration stays stable across all providers.

Let me know which one you think is the higher priority, and I’ll get started....

KERDAWY-2:
Hi, I'm Abdelrahman, a GSoC 2026 applicant interested in **Idea #2 – Multimodal AI and Agent API Eval Framework**.

While exploring the codebase as part of my pre-application steps, I came across this meta issue and the 8 sub-issues tracked here. I'd like to understand if any of these sub-issues are currently unassigned and open for contribution, particularly anything related to the AI model selector UI/UX that could help me get familiar with the relevant parts of the codebase.

I have a background in Python, Flutter/Dart, and ML, so I'm comfortable working on both the UI and backend logic. Happy to start with a smaller scoped task here if that's helpful.

Thanks!

TheAnshulPrakash:
Hi @KERDAWY-2 ,

Yes, you can propose a fix for this issue.

[...]

PR #1463, opened 2026-03-25
Fix empty text of Select Model button when no model is chosen

tech-nitin:
Fixes an issue where the Select Model button displays empty text when no model is selected.

Previously, an empty string was shown instead of fallback text.

Now it correctly displays "Select Model" when no valid model is selected, improving usability.

### DIFF
--- a/lib/screens/common_widgets/ai/ai_model_selector_button.dart
+++ b/lib/screens/common_widgets/ai/ai_model_selector_button.dart
@@ -23,34 +23,39 @@ class AIModelSelectorButton extends StatelessWidget {
   });
 
   @override
-  Widget build(BuildContext context) {
-    return ElevatedButton(
-      onPressed: readonly
-          ? null
-          : () async {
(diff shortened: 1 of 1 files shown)


TheAnshulPrakash:
Hey @tech-nitin,

Could you add a bit more detail about what changes were made in the code to fix this issue? The current description explains the result, but it would help reviewers understand the implementation better.

Also, it would be great if you could align the description with the usual PR format we follow and reference the issue you resolved (which I believe is issue #942 ).

Lastly, could you attach a short demo (GIF/video/image) showing the behavior before and after the fix? That would make it easier to verify the improvement during review.

Also check issue #1269 where the maintainers have asked for the complete revamp of the AI model selector

TheAnshulPrakash:
Also, I think you might have opened two PRs addressing the same issue by mistake.
#1464

tech-nitin:
Closing this PR as it is a duplicate. Continuing the work in #1464.

PR #1464, opened 2026-03-25
Fix empty text of Select Model button when no model is chosen

### LINKED ISSUE #942: Empty text of Select model button when no model is chosen.

tech-nitin:
## Description

Fixes an issue where the "Select Model" button displays empty text when no model is selected.

## Changes
- Added a check to handle both null and empty string values for `aiRequestModel.model`
- Introduced fallback text (`kLabelSelectModel`) when no valid model is selected
- Updated logic in `AIModelSelectorButton` to prevent rendering empty text

## Result
- Before: Button displayed empty text ❌
- After: Button displays "Select Model" when no model is selected ✅

Fixes #942


tech-nitin:
Thanks for the feedback!

I’ve updated the PR description with more details about the implementation.

### Changes made:
- Added a check to handle both null and empty string cases for the selected model
- Ensured fallback text "Select Model" is displayed when no valid model is selected
- Updated the UI logic in AIModelSelectorButton to improve user experience

Fixes #942

I will also attach a short demo showing the behavior before and after the fix.

tech-nitin:
## Demo

### After Fix ✅
- The "Select Model" button is now visible when no model is selected
- Clicking it opens the model/provider selection dialog correctly

demo

animator:
superseded by #1779

Issue #1283, opened 2026-03-06
Typo in Request Body: "answer" displayed as "anwe\nr" (shifted r)


alaotach:
The request body field **`answer`** is incorrectly rendered as **`anwe`**, where the character **`r`** appears below the word instead of being placed correctly at the end. This causes the parameter name to appear malformed in the request body.

---

#### Steps to Reproduce the bug/problem

1. Open the **request body** section where parameters are defined.
2. Locate the field intended to be **`answer`**.
3. Observe that the text appears as **`anwe`**, with the **`r`** positioned below the word instead of inline.

---

#### Expected behavior

The parameter name should appear correctly as **`answer`**, with all characters displayed inline in the correct order.

---

#### Device Info (The device where you encountered this issue)

* **OS:** Windows
* **Version:** Windows 11 (25H2)
* **Browser:** Firefox

---

#### Flutter Doctor

```
Doctor summary (to see all details, run flutter doctor -v):

[√] Flutter (Channel stable, 3.41.4, on Microsoft Windows [Version 10.0.26200.7840], locale en-US)
[√] Windows Version (Windows 11 or higher, 25H2, 2009)

[!] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
    X cmdline-tools component is missing.
      Try installing or updating Android Studio.
      Alternatively, download the tools from:
      https://developer.android.com/studio#command-line-tools-only
      and make sure to set the ANDROID_HOME environment variable.
      See https://developer.android.com/studio/command-line for more details.

    X Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/to/windows-android-setup for more details.

[√] Chrome - develop for the web
[√] Visual Studio - develop Windows apps (Visual Studio Community 2026 18.0.2)
[√] Connected device (3 available)
[√] Network resources

! Doctor found issues in 1 category.
```

pratham-sarankar:
@alaotach 

I tried reproducing this issue on the latest codebase but wasn’t able to observe the problem anymore.

Tested on:
macOS (Apple Silicon)
Flutter 3.x / Dart 3.x

tech-nitin:
Hi, I’d like to work on this issue. Can you assign it to me?

alaotach:
@tech-nitin please read the contribution guidelines first, and the issues here are not assigned to anyone you're free to solve any issue and submit a pr for it and the best pr will be accepted

tech-nitin:
[...]

PR #1478, opened 2026-03-26
fix(json_field_editor): pass correct style for special characters

vinimlo:
## Summary
Fix copy-paste error in json_field_editor where `specialCharHighlightStyle` was receiving `stringHighlightStyle` instead of the correct field defined in the widget state.

Special characters (`{`, `}`, `[`, `]`, `,`, `:`) were being styled using the string highlight style instead of their own `specialCharHighlightStyle` (a grey color defined on line 164-166 of `json_field.dart`).

This has been like this since the first commit of json_field_editor — the field and default style were set up correctly, but the wiring to `JsonHighlight` used the wrong variable.

## Changes
One-line fix in `packages/json_field_editor/lib/src/json_field.dart` line 244.

## Test plan
- [x] All 11 existing `json_field_editor` tests pass (`flutter test`)
- [x] Verified on macOS
- [ ] Could not test on Windows (no access to a Windows machine)

Related: #1283

### DIFF
--- a/packages/json_field_editor/lib/src/json_field.dart
+++ b/packages/json_field_editor/lib/src/json_field.dart
@@ -241,7 +241,7 @@ class JsonFieldState extends State<JsonField> {
             keyHighlightStyle: keyHighlightStyle,
             nullHighlightStyle: nullHighlightStyle,
             numberHighlightStyle: numberHighlightStyle,
-            specialCharHighlightStyle: stringHighlightStyle,
+            specialCharHighlightStyle: specialCharHighlightStyle,
             stringHighlightStyle: stringHighlightStyle,
             commonTextStyle: commonTextStyle,
             isFormating: widget.isFormatting,


vinimlo:
Just to add some context here — I traced this back to the very first commit of json_field_editor (474b9d71). The field specialCharHighlightStyle was properly defined with a grey[700] default and exposed as a widget parameter, but it was never actually wired into the JsonHighlight builder. It looks like a copy-paste oversight during the initial implementation.

Since this has been shipping this way from day one, special characters ({, }, [, ], commas, colons) have always been styled using the string color instead of their intended grey. This fix will change their visual appearance to match what was originally designed.

I ran all 11 json_field_editor tests and they pass. I was able to verify the fix on macOS but could not test on Windows since I don't have access to a Windows machine — would appreciate if a reviewer could check that as well.

PR #1479, opened 2026-03-26
Fix Select Model button fallback text for empty model string

wxxz123:
## PR Description

This PR fixes a bug where the "Select Model" button displays an empty text instead of its fallback label. 
Previously, the button used the `??` operator which only checks for strict `null`. However, when a selected model string is cleared or evaluates to an empty string `""`, the text widget would render a blank space instead of showing `kLabelSelectModel`. This PR updates the logic to explicitly check `aiRequestModel?.model?.isNotEmpty` to ensure the fallback text is always displayed correctly.

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: This is a trivial UI fallback text fix that does not affect core logic, so specific unit tests are not necessary.

- [x] macOS
- [ ] Windows
- [ ] Linux


animator:
superseded by #1779

PR #1522, opened 2026-03-27
Fix empty model text

tech-nitin:
Fixes issue where the "Select Model" button displayed empty or unclear text when no model was selected.

## Problem
The button relied only on null checks, which caused empty strings to be rendered, resulting in unclear UI.

## Solution
- Added proper fallback handling for both null and empty model values
- Ensured "Select Model" is displayed when no valid model is selected

## Result
- Improved clarity of the UI
- Better user experience in model selection

This aligns with my GSoC proposal focused on improving AI model handling and user experience.


animator:
Proposal PR should not have any code changes.

PR #1619, opened 2026-03-31
fix(request): stop sendRequest early when validation fails

AyushCoder9:
## PR Description

This PR ensures `sendRequest()` exits immediately when request validation fails.

Before this change, validation errors were logged in terminal but execution could still continue and start the network flow. Now, after logging the validation error and setting terminal badge, the function returns early. This keeps behavior safe and avoids unnecessary network calls for invalid requests.

Also added a regression test:
- `test/providers/collection_validation_test.dart`
  - verifies `sendRequest()` early-exits on invalid request input
  - verifies validation error is logged in terminal
  - verifies no response status is set from network execution.

## Related Issues

- Closes #

### Checklist

## Added/updated tests?

- [x] Yes

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/providers/collection_providers.dart
+++ b/lib/providers/collection_providers.dart
@@ -377,6 +377,7 @@ class CollectionStateNotifier
         level: TerminalLevel.error,
       );
       ref.read(showTerminalBadgeProvider.notifier).state = true;
+      return;
     }
 
     // Terminal: start network log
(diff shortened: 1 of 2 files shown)


animator:
The purpose is for terminal logging so it is fine, let the server send back the error.

PR #1638, opened 2026-04-06
addedSniffingLogicInModelLayer

### LINKED ISSUE #1382: Previewer fallback based on magic bytes

hihry:
Fixes #1382 
## Summary
Moves response content sniffing and preview-type selection into the model/networking layer so content interpretation happens once, before any UI rendering.

## Problem
Through sniffing the magic bytes we are routing the application/stream-octet to a appropriate previewer

## Solution
- Response sniffing now happens in the model/networking flow
- Content type is classified once and exposed as structured model data
- Improved content-type utility handling for ambiguous or missing headers

This should be architecturally better approach 

## Why Model Layer
| Concern | Before | After |
|---|---|---|
| Logic location | Widget layer | Model/networking layer |
| Source of truth | Multiple widgets | Single classification |
| Testability | Widget-level tests | Pure unit tests |
| Reusability | Per-view | Desktop, mobile, web |
| Rebuild performance | Repeated parsing | Computed once |

## Impact
- No UI changes
- Reduces renderer mismatch risk across all response consumers

## Demo

https://github.com/user-attachments/assets/cb69823b-0861-4358-b971-8d2bbab14f92

### DIFF
--- a/packages/better_networking/lib/models/http_response_model.dart
+++ b/packages/better_networking/lib/models/http_response_model.dart
@@ -60,7 +60,7 @@ class HttpResponseModel with _$HttpResponseModel {
       _$HttpResponseModelFromJson(json);
 
   String? get contentType => headers?.getValueContentType();
-  MediaType? get mediaType => getMediaTypeFromHeaders(headers);
+  MediaType? get mediaType => getMediaTypeFromHeadersOrSniff(headers, bodyBytes);
 
   HttpResponseModel fromResponse({
     required Response response,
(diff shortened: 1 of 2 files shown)


animator:
Do not reinvent the wheel. A simpler solution exists.

hihry:
Hi, thanks for the review! I want to make sure I'm heading in the right direction before reworking this.
My current approach manually sniffs MIME types via magic bytes in `content_types_utils.dart` which is in packages layer instead of in widget layer. When you mention "a simple solution exists" — could you give me a small nudge on what direction to look? For instance:

There is one way of doing this through external package, but it has it's own pitfalls.
or something already present within the apidash codebase that I may have missed?

I want to avoid submitting another PR that misses the mark. Even a one-line hint would go a long way — I'll take it from there. Thanks!
@animator

PR #1646, opened 2026-04-07
fix: escape special characters in environment variable keys for regex substitution

### LINKED ISSUE #1645: [Bug]: Environment variable keys with special characters crash app with FormatException
## Describe the bug

The `substituteVariables` function in `lib/utils/envvar_utils.dart` builds a regex from **all** active environment variable keys without escaping regex metacharacters. If any key contains an unclosed metacharacter like `(` or `[`, a `FormatException` is thrown before the request is sent — **silently freezing all requests** in that environment with no error shown to the user.

## Steps to Reproduce

1. Go to **Environments** → create a new environment
2. Add a variable with key `api(key` and any value (the `(` makes the regex malformed)
3. Set it as the **active environment**
4. Send **any** request — even a plain GET with no variables in the URL
5. **Result:** The request never sends, no response, no error message. All subsequent requests in that environment are also frozen.

*(See attached screen recording)*

## Root Cause

**File:** `lib/utils/envvar_utils.dart`, line 47

```dart
// Buggy — all keys joined raw into regex, no escaping
final regex = RegExp("{{(${envVarMap.keys.join('|')})}}");
// Key "api(key" produces: {{(api(key)}}
// → FormatException: Unterminated group (thrown silently before isWorking is set)
```

Because the exception fires before `isWorking: true` is set, the Send button stays enabled but nothing ever executes — making it look like the button is broken with no feedback.

## Why It Affects All Requests

The regex is built from **all** keys in the environment map on every request, not just the ones referenced in the URL. One bad key corrupts the regex for every request.

## Proposed Fix

```dart
final escapedKeys = envVarMap.keys.map((key) => RegExp.escape(key)).join('|');
final regex = RegExp("{{($escapedKeys)}}");
```

## Impact

- **Severity: High** — a single env var key with `(` or `[` silently disables all requests in that environment
- **Platforms:** All (cross-platform)
- **Additional issue:** Keys with `.`, `+`, `*` don't crash but produce wrong regex — variables are not substituted correctly

Fikri-20:
## Description

This PR fixes a critical bug where environment variable keys containing regex metacharacters caused the application to crash with a `FormatException`.

The `substituteVariables` function in `lib/utils/envvar_utils.dart` was building regex patterns by joining environment variable keys without escaping special characters. This meant that keys like `api.key`, `user+id`, `test[0]`, etc. would either create incorrect regex patterns or throw exceptions during pattern compilation.

### 1. Core Fix (`lib/utils/envvar_utils.dart`)
- Added `RegExp.escape()` to properly escape all special characters in environment variable keys before building the regex pattern
- Changed from: `final regex = RegExp("{{(${envVarMap.keys.join('|')})}}");`
- To: `final escapedKeys = envVarMap.keys.map((key) => RegExp.escape(key)).join('|'); final regex = RegExp("{{($escapedKeys)}}");`

### 2. Test Coverage (`test/utils/envvar_utils_test.dart`)
Added comprehensive test cases to verify the fix handles:
- **Special characters**: `api.key`, `user+id`, `var*name`, `test[0]`
- **Regex metacharacters**: `api$key`, `var^name`, `test|pipe`, `back\slash`
- **Mixed scenarios**: Normal variables combined with special character variables

## Testing

All existing tests continue to pass, and new tests verify the fix works correctly for edge cases.

### Test Results
- ✅ All original test cases pass
- ✅ New test case: Special characters in variable keys
- ✅ New test case: Regex metacharacters in variable keys  
- ✅ New test case: Mixed special and normal characters

## Related Issue

Fixes #1645

## Impact

**Before:** Users would experience app crashes when using environment variables with special characters in their names, which is a common naming convention (e.g., dot notation like `api.key`).

**After:** Environment variables now support all common naming conventions including dots, dashes, plus signs, brackets, and other special characters without any issues.

## Checklist

- [x] Bug fix implemented with proper escaping
- [x] Comprehensive test cases added
- [x] All existing tests continue to pass
- [x] Code follows project conventions
- [x] No breaking changes introduced

### DIFF
--- a/lib/utils/envvar_utils.dart
+++ b/lib/utils/envvar_utils.dart
@@ -44,7 +44,8 @@ String? substituteVariables(
   if (envVarMap.keys.isEmpty) {
     return input;
   }
-  final regex = RegExp("{{(${envVarMap.keys.join('|')})}}");
+  final escapedKeys = envVarMap.keys.map((key) => RegExp.escape(key)).join('|');
+  final regex = RegExp("{{($escapedKeys)}}");
 
   String result = input.replaceAllMapped(regex, (match) {
     final key = match.group(1)?.trim() ?? '';
--- a/test/utils/envvar_utils_test.dart
+++ b/test/utils/envvar_utils_test.dart
@@ -168,6 +168,41 @@ void main() {
       String expected = "{{url1}}/humanize/social?num=8940000";
       expect(substituteVariables(input, combinedEnvVarsMap), expected);
     });
+
+    test("Testing substituteVariables with special characters in variable keys", () {
+      Map<String, String> envMap = {
+        "api.key": "value_with_dot",
+        "user+id": "value_with_plus",
+        "var*name": "value_with_asterisk",
+        "test[0]": "value_with_brackets",
+      };
+      String input = "API: {{api.key}}, User: {{user+id}}, Var: {{var*name}}, Test: {{test[0]}}";
+      String expected = "API: value_with_dot, User: value_with_plus, Var: value_with_asterisk, Test: value_with_brackets";
+      expect(substituteVariables(input, envMap), expected);
+    });
+
+    test("Testing substituteVariables with regex metacharacters in keys", () {
+      Map<String, String> envMap = {
+        "api\$key": "dollar_value",
+        "var^name": "caret_value",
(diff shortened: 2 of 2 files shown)


animator:
Crash video is not added.

PR #1688, opened 2026-05-30
fix: redact sensitive oauth diagnostics

### LINKED ISSUE #1686: Advanced Security: Prevent sensitive data exposure in logging mechanisms
### Description
When making API requests, authorization headers or sensitive payload data can inadvertently be printed to the console or logs during development/debugging, which risks leaking secrets if a crash report is generated.

### Proposed Solution
I propose implementing a data sanitization layer in the network interceptors that automatically redacts `Authorization`, `Cookie`, and specific sensitive JSON keys (like passwords) before any network payload is logged or recorded by the app.

**I would love to work on this advanced security issue under the GSSoC '24 program! Could you please assign it to me?**

saurabhhhcodes:
## Summary
- redact OAuth access tokens before debug logging in auth handling
- redact sensitive OAuth callback query parameters before logging callback requests
- add unit coverage for the shared redaction helper

## Validation
- git diff --check
- Could not run Dart/Flutter tests locally because dart and flutter are not installed in this environment.

Fixes #1686

### DIFF
--- a/packages/better_networking/lib/services/oauth_callback_server.dart
+++ b/packages/better_networking/lib/services/oauth_callback_server.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 import 'dart:async';
 import 'dart:developer' show log;
+import 'package:better_networking/utils/sensitive_data_redactor.dart';
 
 /// A lightweight HTTP server for handling OAuth callbacks on desktop platforms.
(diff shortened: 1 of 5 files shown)


apidash-dev:
Thanks for the PR.
debugPrint triggers only while in testing and not in release.
So not an issue.

PR #1695, opened 2026-06-06
Shashwat/storage migration

### LINKED ISSUE #489: Sharing Collection in JSON file formate
I was using the application and I realized that there isn't a feature to share the whole collection that I can share within my team which makes it very easy for other developers to understand the working of API without the hassle of creating new collection manually

We can have a feature where user can generate and export JSON file of any collection which anyone can import to setup the shared collection

Also we can add more into importing by making it flexible for url, raw JSON file or JSON in text. And also giving an option to export JSON on the basis of applications like postman which will make it easy for devs who use postman to share the collection between apidash and postman

ShashwatXD:
Storage migration to file system.

closes #489

## Added

* Filesystem-based workspaces (`collections/`, `environments/`, `history/` as JSON)
* Multi-collection catalog with active collection providers
* Autosave support (removed need for manual saves)
* Workspace picker with recent workspaces
* Secure storage for secrets and AI keys
* Media response bodies stored as sidecar files
* Storage and workspace tests

## Removed

* Hive persistence (`hive_services`)
* Sidebar **Save** button (and its associated test)


ShashwatXD:
@animator  Please review this PR.

lib/services/storage/workspace_storage.dart
```diff
@@ -0,0 +1,828 @@
...
+    final requestJson = _readJsonSync(_requestJsonRelative(collectionId, id));
+    if (requestJson == null) {
+      return null;
+    }
+    final merged = Map<String, Object?>.from(requestJson);
+    final responseJson = _readJsonSync(_responseJsonRelative(collectionId, id));
+    if (responseJson != null) {
+      final requestDirPath = _path(_requestDirRelative(collectionId, id));
+      merged['httpResponseModel'] =
+          _inlineResponseBodyFile(requestDirPath, responseJson);
```
synapsecode, line 364: are we doing any validations on the filenames if they're coming from the responseJson directly?
if these workspaces can be imported and shared via git and what not then certain relative paths and stuff might be able to cause unintended effects. 
lmk if this is handled

pubspec.yaml
```diff
@@ -44,10 +44,10 @@ dependencies:
   flutter_portal: ^1.1.4
   flutter_riverpod: ^3.2.1
   flutter_svg: ^2.2.4
+  flutter_secure_storage: ^9.2.4
   fvp: ^0.35.2
   highlight: ^0.7.0
   highlighter: ^0.1.1
-  hive_ce_flutter: ^2.3.4
```
synapsecode, line 50: minor: pubspec.lock still contains 
```
 hive_ce_flutter:
    dependency: "direct main"
    description:
      name: hive_ce_flutter
      sha256: "2677e95a333ff15af43ccd06af7eb7abbf1a4f154ea071997f3de4346cae913a"
      url: "https://pub.dev"
    source: hosted
    version: "2.3.4"
```
maybe you missed doing pub get and then adding lockfile also to changes?

lib/screens/common_widgets/sidebar_header.dart
```diff
@@ -23,8 +23,11 @@ class SidebarHeader extends ConsumerWidget {
       padding: kPe8,
       child: Row(
         children: [
-          const SaveButton(),
-          const Spacer(),
+          if (kIsDesktop)
+            const Expanded(
+              child: WorkspaceDropdown(),
+            ),
+          if (kIsDesktop) kHSpacer4,
```
synapsecode: nitpick: could be put in a spread ... [...] i guess

lib/services/storage/workspace_paths.dart
```diff
@@ -0,0 +1,17 @@
...
+Future<String?> resolveWorkspaceRoot({
+  required bool useDesktopPath,
+  String? desktopPath,
+}) async {
+  if (useDesktopPath) {
+    if (desktopPath == null || desktopPath.isEmpty) {
+      return null;
+    }
+    return desktopPath;
+  }
```
synapsecode, line 14: why do we have useDesktopPaht and desktopPath both?
like if desktopPath is non-null cant we assume that we must use it?

lib/screens/envvar/environment_page.dart
```diff
@@ -12,6 +12,10 @@ import 'environment_editor.dart';
 class EnvironmentPage extends ConsumerWidget {
   const EnvironmentPage({super.key});
 
+  static bool _shouldShowEnvironmentMenu(String environmentId) {
+    return environmentId != kGlobalEnvironmentId;
+  }
```
synapsecode: nitpick: could have been inlined, no need for separate function

lib/services/storage/workspace_storage.dart
ShashwatXD: thanks for flagging this, it only affected media responses saved as sidecar files (png/pdf/etc.). Fixed it.

lib/services/storage/workspace_paths.dart
ShashwatXD: useDesktopPath tells us whether we're in desktop or mobile mode, while desktopPath is just the selected folder.
Since this function is used by both platforms:

- Desktop + no path yet-  return null to show the workspace picker and if path is not null then it opens that workspace.
-  Mobile -ignore desktopPath and use the app documents directory.

pubspec.yaml
ShashwatXD: updated.

PR #1700, opened 2026-06-16
GSoC2026/Borgy: MCP App tools experiment

AbdelrahmanELBORGY:
# Description - API Dash MCP App
This PR contains the MCP server and basic MCP app UI for API Dash. 
It allows AI assistants inside editors like VS Code (e.g., Claude Code, Roo Code) to interact directly with the API Dash core engine, execute API requests, and view rich, interactive UIs directly within the editor.

# Visual view

| Execute Request | List History |
| :---: | :---: |
|requestExecution|listHistory|

# 1. VS Code MCP Server Setup

**⚠️ Prerequisites:** 
1. **Node.js:** You must have **[Node.js](https://nodejs.org/)** installed on your system to run the MCP server wrapper.
2. **Build the Desktop App:** 
* Because this headless engine relies on API Dash's Flutter native plugins (like local database and networking), you must compile the app for your desktop operating system at least once. 
* Run `flutter build windows --release` (or `macos` / `linux`) in the root directory to generate the required native build files before starting the server.

To connect this engine to your VS Code AI assistant (like Roo Code or Cline), you must create a `.vscode` folder in the root of your workspace, and then create an `mcp.json` file inside it to register the server.

Add the following configuration to your new `.vscode/mcp.json` file, making sure to replace `<PATH_TO_APIDASH_REPO>` with the actual absolute path to where you cloned this repository on your machine:

```json
{
  "servers": {
    "apidash-mcp": {
      "command": "node",
      "args": [
        "<PATH_TO_APIDASH_REPO>/scripts/apidash-mcp.js"
      ]
    }
  }
}
```

**Examples of valid paths:**
* **Windows:** `"C:/Users/Username/Documents/apidash/scripts/apidash-mcp.js"`
* **macOS/Linux:** `"/Users/Username/projects/apidash/scripts/apidash-mcp.js"`

*(Note: Use standard forward slashes `/` in the JSON path, even on Windows, as Node.js handles them perfectly and it prevents escaping errors).*

---

# 2. Architecture and File Overview

### Added Files
*   **`mcp_engine.dart`**: Acts as the core JSON-RPC protocol router. It listens for incoming tool executions or resource requests, matches the requested method (e.g., `tools/call`), and dispatches it to the appropriate handler while catching errors to prevent server hangs.
*   **`mcp_service.dart`**: Manages the lifecycle of the headless Dart process. It bridges standard input/output (stdio) to the isolate ports, allowing the underlying Flutter/Dart engine to communicate with the Node.js wrapper.
*   **`tool_handler.dart`**: The brain of the operations. It registers the schemas for the AI tools and executes the actual backend logic (like sending HTTP requests and saving to the database).
*   **`resource_handler.dart`**: Responsible for providing the MCP client with the correct UI template based on the requested `ui://` URI scheme.
*   **`html_templates.dart`**: Contains the raw HTML, CSS, and Javascript needed to render the API Dash dashboards inside VS Code's webview panels.

### The 4 Core Tools and How They Invoke Each Other
1.  **`apidash_execute_request`**: 
    *   *Function*: The AI calls this tool to construct and fire off an HTTP request using the API Dash core. 
    *   *Invocation*: Once the request completes, it saves the data to the local database, generates a unique Execution ID, and returns a `_meta` UI trigger. This forces VS Code to open the Results Dashboard.
2.  **`apidash_get_results`**: 
    *   *Function*: This tool acts as the data bridge for the UI. It is explicitly hidden from the AI agent using the `visibility: ["app"]` modifier in its `_meta` schema.
    *   *Invocation*: The AI cannot see or call this tool. Instead, the Javascript running inside the Results Dashboard UI polls this tool continuously to fetch the latest request data from the local database and update the screen.
3.  **`apidash_list_history`**: 
    *   *Function*: Fetches all past requests from the local database.
    *   *Invocation*: When the AI calls this, the Dart backend returns the history list as `structuredContent` (raw JSON) and sends a `_meta` trigger to open the History UI. The UI's Javascript automatically intercepts this raw JSON to render the interactive list.
4.  **`apidash_delete_request`**: 
    *   *Function*: Deletes a specific request from the database. It is explicitly hidden from the AI agent using the `visibility: ["app"]` modifier.
    *   *Invocation*: This is invoked silently by the History UI. When a user clicks the "Delete" button on the screen, the Javascript uses `window.parent.postMessage` to secretly ask the MCP client to call this tool on the backend, bypassing the AI entirely.

# Invoking visible to app only tools - not visible to the ai agent

|`apidash_get_results`|`apidash_delete_request`|
| :---: | :---: |
|executeRequest|deleteRequest|

---

## 3. The Node.js Wrapper & Stdio Communication

You might wonder why a `.js` file is used as the entry point when the entire application is built in Dart/Flutter. 

**Why a `.js` script?**
Most MCP clients running inside VS Code expect to spawn a server process via a highly standard environment (like Node.js or Python) and communicate over standard Input/Output (`stdio`). The Node.js script acts as a lightweight, reliable proxy wrapper.

**Why this approach is necessary:**
API Dash relies heavily on Flutter-specific packages (like `WidgetsFlutterBinding`) and complex local databases. If you try to run a pure, raw Dart script from the command line without the proper Flutter context, these core packages will crash. 

By starting a Flutter-based headless app, we initialize all the necessary core packages securely. The Node.js script simply pipes standard JSON-RPC text strings via `stdin` into the headless Dart environment, and the Dart environment pushes responses back out via `stdout`. The `McpService` class actively listens to this `stdin` stream, processes the JSON, and pipes it into the isolated MCP engine.

---

## 4. Hive Boxes for Local Storage

The engine utilizes Hive CE (`hive_ce`) as its local storage mechanism.

**What it does:**
Initialize an isolated `agent_history` box specifically for the MCP server.

**Why it is important:**

* **Persistence:** Unlike keeping data in RAM (which vanishes the moment the AI finishes its thought process), Hive permanently saves every request to the physical hard drive.
* **Stateless UI Synchronization:** Because the AI and the UI operate completely asynchronously, Hive acts as the central source of truth. The AI saves the request data to Hive and calls `.flush()` to guarantee it is written to the disk. Milliseconds later, the UI wakes up, queries Hive, and is guaranteed to find the data.
* **Historical Records:** By generating a unique ID (`execution_id`) for every request, Hive allows us to build a comprehensive, permanent history log that the user can browse and manage.

### DIFF
--- a/lib/mcp/resources/html_templates.dart
+++ b/lib/mcp/resources/html_templates.dart
@@ -1,0 +1,282 @@
+String getApiDashAppHtml() {
+  return r'''
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>API Dash Execution Dashboard</title>
+    <style>
+        * { box-sizing: border-box; margin: 0; padding: 0; }
+        body { font-family: system-ui, -apple-system, sans-serif; background: var(--host-bg, #0f0f0f); color: var(--host-color, #e0e0e0); padding: 20px; min-height: 400px; }
+        .topbar { display: flex; align-items: center; justify-content: space-between; margin-bottom: 20px; }
+        .topbar h1 { font-size: 15px; font-weight: 600; color: #a78bfa; display: flex; align-items: center; gap: 8px; }
+        .badge { font-size: 11px; font-weight: 600; padding: 3px 9px; border-radius: 20px; }
+        .badge.ok  { background: #14532d; color: #86efac; }
+        .badge.err { background: #450a0a; color: #fca5a5; }
+        .id-badge { font-size: 10px; background: #2a2a2a; color: #9ca3af; padding: 2px 6px; border-radius: 4px; font-family: monospace; font-weight: normal; margin-left: 8px; }
+        .grid { display: grid; grid-template-columns: repeat(3, 1fr); gap: 12px; margin-bottom: 16px; }
+        .card { background: #1a1a1a; border: 1px solid #2a2a2a; border-radius: 10px; padding: 14px 16px; }
+        .card .label { font-size: 11px; color: #6b7280; text-transform: uppercase; margin-bottom: 6px; }
+        .card .value { font-size: 22px; font-weight: 600; color: #e5e7eb; }
+        .section { background: #1a1a1a; border: 1px solid #2a2a2a; border-radius: 10px; padding: 16px; margin-bottom: 12px; }
+        .section-title { font-size: 12px; font-weight: 600; color: #6b7280; text-transform: uppercase; margin-bottom: 12px; }
+        .url-row { display: flex; align-items: center; gap: 10px; padding: 10px 12px; background: #111; border-radius: 8px; border: 1px solid #2a2a2a; }
+        .method-pill { font-size: 11px; font-weight: 700; padding: 3px 8px; border-radius: 6px; background: #1e3a5f; color: #60a5fa; }
+        .url-text { font-size: 13px; color: #9ca3af; font-family: monospace; word-break: break-all; }
+        .body-pre { background: #111; border-radius: 8px; padding: 14px; font-size: 12px; font-family: monospace; color: #d1d5db; overflow-x: auto; white-space: pre-wrap; word-break: break-all; border: 1px solid #2a2a2a; max-height: 240px; overflow-y: auto;}
+    </style>
+</head>
+<body>
+
+<div class="topbar">
+    <h1>⚡ API Dash <span class="id-badge" id="req-id">Loading...</span></h1>
+    <span class="badge ok" id="status-badge" style="background:#374151; color:#9ca3af;">Loading Data...</span>
+</div>
+
+<div class="grid">
+    <div class="card"><div class="label">Status</div><div class="value" id="stat-status">—</div></div>
+    <div class="card"><div class="label">Time</div><div class="value" id="stat-time">—</div></div>
+    <div class="card"><div class="label">Size</div><div class="value" id="stat-size">—</div></div>
+</div>
+
+<div class="section">
+    <div class="section-title">🔗 Request</div>
+    <div class="url-row">
+        <span class="method-pill" id="req-method">—</span>
+        <span class="url-text" id="req-url">—</span>
+    </div>
+</div>
+
+<div class="section">
+    <div class="section-title">📄 Response body</div>
+    <pre class="body-pre" id="body-pre">Fetching from backend engine...</pre>
+</div>
+
+<script>
+    let currentDisplayedId = null;
+
+    function hydrate(data) {
+        if (!data) return;
+        // Skip updating the UI if the execution ID hasn't changed
+        if (currentDisplayedId === data.execution_id) return;
+        currentDisplayedId = data.execution_id;
+
+        const code = data.status_code || 0;
+        const ok = code >= 200 && code < 300;
+
+        if (data.execution_id) {
+            document.getElementById('req-id').textContent = data.execution_id;
+        }
+
+        document.getElementById('stat-status').textContent = code;
+        document.getElementById('stat-status').style.color = ok ? '#4ade80' : '#f87171';
+        document.getElementById('stat-time').textContent = (data.time_ms || 0) + ' ms';
+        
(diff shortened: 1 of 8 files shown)


scripts/apidash-mcp.js
animator, line 1: How will this file be shipped to the end user?

lib/mcp/resources/html_templates.dart
animator, line 1: Separate files for separate templates.
Why are you not following the best practices that have been already documented in the article provided to you - https://dev.to/ashita/a-practical-guide-to-building-mcp-apps-1bfm

- separate files for UI resources
- shared css design system

Why polling is implemented?

lib/mcp/resources/html_templates.dart
```diff
@@ -0,0 +1,282 @@
...
+
+<div class="topbar">
+    <h1>Request History</h1>
+</div>
+
+<div id="loading" class="loading">Fetching history...</div>
+<div id="history-container" class="history-list"></div>
+
+<script>
+    async function executeMcpTool(name, args) {
```
animator, line 182: Why is this not a common utility? As architected in the article.

lib/mcp/resources/resource_handler.dart
```diff
@@ -0,0 +1,39 @@
+import 'html_templates.dart';
+
+class ResourceHandler {
+  static Map<String, dynamic> listResources() {
```
animator, line 4: Where is this function being used?

lib/mcp/resources/resource_handler.dart
```diff
@@ -0,0 +1,39 @@
...
+        {
+          "uri": "ui://apidash-agentic-engine/dashboard/history",
+          "name": "API Dash Request History",
+          "mimeType": "text/html;profile=mcp-app",
+        }
+      ],
+    };
+  }
+
+  static Map<String, dynamic> readResource(String uri) {
```
animator, line 21: Where is this function being used?

lib/mcp/tools/tool_handler.dart
```diff
@@ -0,0 +1,220 @@
...
+            "required": ["url", "method"],
+          },
+          "_meta": {
+            "ui": {
+              "resourceUri": "ui://apidash-agentic-engine/dashboard/results"
+            }
+          },
+        },
+        {
+          "name": "apidash_get_results",
```
animator, line 32: why is results a separate app?

lib/mcp/resources/resource_handler.dart
AbdelrahmanELBORGY: it tells the IDE client the list of the resources including interfaces that the MCP server has, by returning a JSON map containing a list of resources, each resource has its own name, uri and mimetype, it is not used directly,it runs once when the MCP server first connects to the IDE client

lib/mcp/resources/resource_handler.dart
AbdelrahmanELBORGY: this one when the IDE client actually decides to open a specific resource, it calls this function, passing in the specific uri

lib/mcp/tools/tool_handler.dart
AbdelrahmanELBORGY: actually this is an hidden from agent tool, it is used internally just to fetch the last request results in order to display it in the ui

scripts/apidash-mcp.js
```diff
@@ -0,0 +1,50 @@
...
+
+const { spawn } = require('child_process');
+const path = require('path');
+const readline = require('readline');
+
+// 1. Specify the exact path to your built release executable
+
+// __dirname is "apidash/scripts"
+// '..' takes us up to the root folder "apidash/"
+const EXE_PATH = path.join(__dirname, '..', 'build', 'windows', 'x64', 'runner', 'Release', 'apidash.exe');
```
animator, line 11: hard coded path .. does not work for end user.

pubspec.yaml
```diff
@@ -75,6 +75,8 @@ dependencies:
     git:
       url: https://github.com/google/flutter-desktop-embedding.git
       path: plugins/window_size
+  genui: ^0.9.2
+  google_generative_ai: ^0.4.7
```
animator, line 79: How is this relevant to this PR?

scripts/apidash-mcp.js
animator: Also now working for macos or linux as written in the instruction.

animator:
Closing the PR as upgraded PR raised.

PR #1706, opened 2026-06-25
fix: prevent null crash on empty JWT private key (#1703)

### LINKED ISSUE #1703: bug: Null crash in substituteAuthModel for JWT privateKey
### Description

AuthJwtModel.privateKey is declared as String? (nullable with no default). The refactored code applies the ! null-assertion operator to the result of substituteVariables(), which explicitly returns null when its input is null.

### Steps to reproduce

1. Open `lib/utils/envvar_utils.dart` around line 143
2. Dart throws 'Null check operator used on a null value' at runtime whenever a user sends a request with JWT auth and has not filled in the optional privateKey field.

### Expected behavior

substituteVariables() returns null when input == null (line 43). The original code used ?? jwt.privateKey as a safe fallback, preserving null for nullable fields. The refactor replaced all ?? fallbacks with !, which is only safe for non-nullable fields. should be handled correctly.

### Actual behavior

Dart throws 'Null check operator used on a null value' at runtime whenever a user sends a request with JWT auth and has not filled in the optional privateKey field.

deepak0x:
## PR Description

`AuthJwtModel.privateKey` is nullable (`String?`), but `substituteAuthModel()` in `lib/utils/envvar_utils.dart` force-unwraps the result of `substituteVariables(jwt.privateKey, envVarMap)!`. Since `substituteVariables()` returns `null` for `null` input, using JWT auth without a private key crashes with **"Null check operator used on a null value."**

This restores the safe `?? jwt.privateKey` fallback for `privateKey`. `secret` and `payload` keep `!` since they are `required String` (non-nullable) and cannot be null.

```dart
privateKey: substituteVariables(jwt.privateKey, envVarMap) ?? jwt.privateKey,
```

## Related Issues

- Closes #1703

### Checklist

## Added/updated tests?

- [x] Yes — added "jwt auth with null private key does not crash" regression test in `test/utils/envvar_utils_test.dart`. Verified locally on Flutter 3.44.4 / Dart 3.12.2 — all tests in `test/utils/envvar_utils_test.dart` pass.

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/utils/envvar_utils.dart
+++ b/lib/utils/envvar_utils.dart
@@ -140,7 +140,8 @@ AuthModel? substituteAuthModel(
         return authModel.copyWith(
           jwt: jwt.copyWith(
             secret: substituteVariables(jwt.secret, envVarMap)!,
-            privateKey: substituteVariables(jwt.privateKey, envVarMap)!,
+            privateKey:
+                substituteVariables(jwt.privateKey, envVarMap) ?? jwt.privateKey,
             payload: substituteVariables(jwt.payload, envVarMap)!,
           ),
         );
--- a/test/utils/envvar_utils_test.dart
+++ b/test/utils/envvar_utils_test.dart
@@ -566,6 +566,40 @@ void main() {
       expect(result.authModel?.jwt?.payload, "payload123");
     });
 
+    test("Testing jwt auth with null private key does not crash", () {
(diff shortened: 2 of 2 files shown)


animator (approved):
Thanks for the fix

PR #1708, opened 2026-06-27
fix: guard null content in saveToDownloads to prevent null-check crash (#1704)

### LINKED ISSUE #1704: bug: Null crash when saving response with null content
### Description

saveToDownloads() accepts a nullable `Uint8List? content` parameter but force-unwraps it with `content!` inside the `if (path != null)` block without any null guard.

### Steps to reproduce

1. Open `lib/utils/save_utils.dart` around line 40
2. Throws 'Null check operator used on a null value' at runtime if the caller provides a non-null path (user picks a save location) but content is null (e.g., empty or failed response body).

### Expected behavior

The null-check only guards the outer `if (path != null)` block, but `content` is a separate nullable that can independently be null. Force-unwrapping it causes an unhandled exception that is not caught by the surrounding try/catch (the crash happens at the `!` dereference before `saveFile` is even called, propagating out of the try block). should be handled correctly.

### Actual behavior

Throws 'Null check operator used on a null value' at runtime if the caller provides a non-null path (user picks a save location) but content is null (e.g., empty or failed response body).

deepak0x:
## PR Description

`saveToDownloads()` in `lib/utils/save_utils.dart` accepted a nullable `Uint8List? content` but force-unwrapped it with `content!` inside the `if (path != null)` block. When a save path is resolved (user picks a location) but `content` is `null` (e.g. an empty or failed response body), this threw `Null check operator used on a null value`.

This PR removes the `content!` force-unwrap and adds an explicit null guard: when `content` is `null`, the user now sees a "No content available to save." snackbar instead of an unhandled crash. Behaviour for valid content and for an unresolved path is unchanged.

## Related Issues

- Closes #1704

### Checklist

## Added/updated tests?

- [x] Yes

Added a regression test in `test/utils/save_utils_test.dart` ("saveToDownloads handles null content without crashing") that calls `saveToDownloads` with `content: null` and asserts a snackbar is shown rather than an exception being thrown.

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/utils/save_utils.dart
+++ b/lib/utils/save_utils.dart
@@ -31,18 +31,19 @@ Future<void> saveToDownloads(
   String? name,
 }) async {
   var message = "";
-  var path = await getFileDownloadpath(
-    name,
-    ext ?? getFileExtension(mimeType),
-  );
+  var path = await getFileDownloadpath(name, ext ?? getFileExtension(mimeType));
   if (path != null) {
-    try {
-      await saveFile(path, content!);
-      var sp = getShortPath(path);
-      message = 'Saved to $sp';
-    } catch (e) {
-      debugPrint("$e");
-      message = "An error occurred while saving file.";
+    if (content == null) {
+      message = "No content available to save.";
+    } else {
+      try {
+        await saveFile(path, content);
+        var sp = getShortPath(path);
+        message = 'Saved to $sp';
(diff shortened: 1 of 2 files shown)


deepak0x:
@animator raised the fix as requested in #1704. Minimal change + regression test. Ready for review.

PR #1709, opened 2026-06-27
fix: guard missing env key in getVariableStatus to prevent null-check crash (#1699)

### LINKED ISSUE #1699: bug: Null crash in getVariableStatus when activeEnvironmentId absent from envMap
### Description

envMap[activeEnvironmentId]! uses the null-assertion operator on a Map lookup, but Map.operator[] returns null when the key is absent. The guard only checks activeEnvironmentId != null, not whether that key exists in the map.

### Steps to reproduce

1. Open `lib/utils/envvar_utils.dart` around line 292
2. Throws 'Null check operator used on a null value' at runtime if getVariableStatus is called with a non-null activeEnvironmentId that has no entry in envMap.

### Expected behavior

envMap[activeEnvironmentId] returns null when the key is missing; the ! operator converts that null into an exception. The sibling function getEnvironmentTriggerSuggestions (line 263) correctly guards with && envMap[activeEnvironmentId] != null before the same lookup. should be handled correctly.

### Actual behavior

Throws 'Null check operator used on a null value' at runtime if getVariableStatus is called with a non-null activeEnvironmentId that has no entry in envMap.

deepak0x:
## PR Description

`getVariableStatus()` in `lib/utils/envvar_utils.dart` guarded only `activeEnvironmentId != null` before doing `envMap[activeEnvironmentId]!`. In Dart, `Map.operator[]` returns `null` for a missing key, so when `activeEnvironmentId` is non-null but absent from `envMap`, the `!` threw `Null check operator used on a null value`.

This PR adds the same `&& envMap[activeEnvironmentId] != null` guard that the sibling function `getEnvironmentTriggerSuggestions` already uses. When the active environment has no entry in `envMap`, the lookup now falls through to global variables (and ultimately the "unknown" result) instead of crashing. Behaviour when the key exists is unchanged.

## Related Issues

- Closes #1699

### Checklist

## Added/updated tests?

- [x] Yes

Added a regression test in `test/utils/envvar_utils_test.dart` ("Testing getVariableStatus with activeEnvironmentId absent from envMap") that calls `getVariableStatus` with a non-null `activeEnvironmentId` that has no entry in `envMap` and asserts it returns the global match instead of throwing. The full `envvar_utils_test.dart` suite passes locally.

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/utils/envvar_utils.dart
+++ b/lib/utils/envvar_utils.dart
@@ -9,37 +9,42 @@ String getEnvironmentTitle(String? name) {
 }
 
 List<EnvironmentVariableModel> getEnvironmentVariables(
-    EnvironmentModel? environment,
-    {bool removeEmptyModels = false}) {
+  EnvironmentModel? environment, {
+  bool removeEmptyModels = false,
+}) {
   if (environment == null) {
     return [];
   }
   return environment.values
-      .where((element) =>
-          element.type == EnvironmentVariableType.variable &&
-          (removeEmptyModels
-              ? element != kEnvironmentVariableEmptyModel
-              : true))
+      .where(
+        (element) =>
+            element.type == EnvironmentVariableType.variable &&
+            (removeEmptyModels
+                ? element != kEnvironmentVariableEmptyModel
+                : true),
+      )
       .toList();
 }
 
 List<EnvironmentVariableModel> getEnvironmentSecrets(
-    EnvironmentModel? environment,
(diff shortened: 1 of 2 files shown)


deepak0x:
@animator raised the fix as requested in #1699. Mirrors the existing guard in getEnvironmentTriggerSuggestions, plus a regression test. Ready for review.

PR #1713, opened 2026-06-28
feat: add in-app update check (#373)

### LINKED ISSUE #373: [feat] in-app update check

deepak0x:
## PR Description

Adds an in-app update check that runs once on launch.

- On startup, queries the GitHub *latest release* API and compares the tag against the running app version (semver, tolerates a leading `v`).
- When a newer version exists, a badge appears on the Settings nav icon and a tile in Settings shows the new version with **Download** (opens the release page) and **Skip this version** actions.
- Skipped versions are persisted, so the badge stays hidden until an even newer release ships.
- The check fails silently on any network/parse error and is disabled during tests, so it never disturbs the user or test runs.

Implementation:
- `services/update_service.dart` — release fetch + version comparison (pure, injectable `http.Client`).
- `providers/update_providers.dart` — `UpdateNotifier` runs the background check and tracks skip state.
- Settings tile + nav badge wiring; skip persisted via shared preferences.

## Related Issues

- Closes #373

### Checklist

## Added/updated tests?

- [x] Yes

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/lib/consts.dart
+++ b/lib/consts.dart
@@ -516,6 +516,8 @@ const kLabelHistory = "History";
 const kLabelLogs = "Logs";
 const kLabelAbout = "About";
 const kLabelSettings = "Settings";
+const kLabelUpdateAvailable = "Update Available";
+const kLabelSkipUpdate = "Skip this version";
 
 // Settings Page
 const kLabelSwitchThemeMode = "Switch Theme Mode";
--- a/lib/providers/providers.dart
+++ b/lib/providers/providers.dart
@@ -6,3 +6,4 @@ export 'terminal_providers.dart';
 export 'settings_providers.dart';
 export 'ui_providers.dart';
 export 'js_runtime_notifier.dart';
+export 'update_providers.dart';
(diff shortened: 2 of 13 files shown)


deepak0x:
@animator please review when you get a chance

PR #1714, opened 2026-06-29
fix: select previous item correctly when deleting from 2-item collection

### LINKED ISSUE #1710: bug: Deleting second-of-two requests leaves selection null
### Description

The remove() function uses `itemIds.length > 1` to decide whether to select the previous item after deletion. When exactly 2 requests exist and the non-first one is deleted, this condition is false (1 item remains), so newId falls through to null — even though a request still exists.

### Steps to reproduce

1. Open `lib/providers/collection_providers.dart` around line 131
2. After the user deletes the second request from a 2-item collection, the sidebar deselects all items and the main panel goes blank, even though one request remains.

### Expected behavior

The guard `else if (itemIds.length > 1)` is too restrictive. The environment's equivalent method (`removeEnvironment`) correctly uses `else if (idx > 0)` instead, which handles a 1-item remainder correctly. should be handled correctly.

### Actual behavior

After the user deletes the second request from a 2-item collection, the sidebar deselects all items and the main panel goes blank, even though one request remains.

Manasvibansal1:
## PR Description

When exactly 2 requests exist and the non-first one is deleted, 
`itemIds.length > 1` evaluates to false (1 item remains), so `newId` 
falls through to null  leaving the main panel blank even though one 
request still exists.

Fix: Replace `itemIds.length > 1` with `idx > 0`, mirroring the 
existing `removeEnvironment` logic.

## Related Issues

- Closes #1710

### Checklist

## Added/updated tests?

- [ ] Yes
- [x] No, This is a one-line logic fix with no new functionality. Existing tests cover this path.

- [ ] Windows
- [x] macOS
- [ ] Linux

### DIFF
--- a/lib/providers/collection_providers.dart
+++ b/lib/providers/collection_providers.dart
@@ -128,7 +128,7 @@ class CollectionStateNotifier
     String? newId;
     if (idx == 0 && itemIds.isNotEmpty) {
       newId = itemIds[0];
-    } else if (itemIds.length > 1) {
+    } else if (idx > 0) {
       newId = itemIds[idx - 1];
     } else {
       newId = null;


Manasvibansal1:
Hi, I've opened a PR with the fix ,  replacing itemIds.length > 1 with idx > 0 to mirror the removeEnvironment logic. PR: #1714

PR #1736, opened 2026-07-16
Fix parseFormData dropping form-data values containing '='

### LINKED ISSUE #1735: bug: parseFormData silently drops form-data values containing '='
### Description

`parseFormData` in `packages/apidash_core/lib/import_export/har_io.dart` splits each `key=value` pair with `pair.split('=')` (line 119) and then only processes it when `keyValue.length == 2` (line 122). Any form-data value containing a literal `=` character (e.g. base64-encoded tokens like `abc==`, OAuth tokens, checksums, or `a=b` style values) produces a split list with length greater than 2, so the key-value pair is silently discarded.

Affected code (`packages/apidash_core/lib/import_export/har_io.dart`, lines 119-127):

```dart
var keyValue = pair.split('=');

// Ensure the pair contains both key and value
if (keyValue.length == 2) {
  var key = Uri.decodeComponent(keyValue[0]);
  var value = Uri.decodeComponent(keyValue[1]);

  result[key] = value;
}
```

Permalink (pinned to current `main`, commit `e346b51`):
https://github.com/foss42/apidash/blob/e346b5188602e45d904db52b31a743dbb6a06955/packages/apidash_core/lib/import_export/har_io.dart#L119-L127

### Steps to reproduce

1. Open `packages/apidash_core/lib/import_export/har_io.dart` (the `split('=')` at line 119 and the `keyValue.length == 2` check at line 122).
2. Import a HAR file with an `application/x-www-form-urlencoded` body where any value contains `=` (for example `token=YWJjZA==` or `filter=a=b`).
3. Those fields are completely lost from the imported request model. The user sees an incomplete form body with no error or warning.

### Expected behavior

`split('=')` splits on every occurrence of `=`, not just the first. The pair should be split only on the first `=` (e.g. via `indexOf` + `substring`) so the value keeps any remaining `=` characters. All fields should be imported correctly.

### Actual behavior

When importing HAR files with `application/x-www-form-urlencoded` bodies where any value contains `=`, those fields are silently dropped from the imported request model (line 122 skips them). The user sees an incomplete form body with no error or warning.

deepak0x:
## PR Description

`parseFormData` in `har_io.dart` split each `key=value` pair with `pair.split('=')` and only kept it when the result had exactly two elements. Any form-data value containing a literal `=` (base64 tokens ending in `=`/`==`, OAuth tokens, checksums, `a=b` style values) produced a list with more than two elements, so the whole pair was silently dropped when importing `application/x-www-form-urlencoded` HAR bodies.

This splits on the first `=` only using `indexOf`/`substring`, so the key is everything before it and the value keeps any remaining `=` characters intact. Behavior for normal pairs, empty values, and null/empty input is unchanged.

Verified with `dart analyze` (no issues) and the added unit tests covering trailing `=`, mid-value `=`, multiple pairs where one value contains `=`, percent-decoding, empty value, and null/empty input.

## Related Issues

- Closes #1735

## Video Demo

- Not applicable; this is a data-parsing fix. The added unit tests cover the before/after behavior (values with `=` were dropped before, preserved now).

### Checklist

## Added/updated tests?

- [x] Yes

- [ ] Windows
- [ ] macOS
- [x] Linux

### DIFF
--- a/packages/apidash_core/lib/import_export/har_io.dart
+++ b/packages/apidash_core/lib/import_export/har_io.dart
@@ -116,12 +116,12 @@ class HarParserIO {
 
     // Loop through the pairs and split them into keys and values
     for (var pair in pairs) {
-      var keyValue = pair.split('=');
+      var separatorIndex = pair.indexOf('=');
 
       // Ensure the pair contains both key and value
-      if (keyValue.length == 2) {
-        var key = Uri.decodeComponent(keyValue[0]);
-        var value = Uri.decodeComponent(keyValue[1]);
+      if (separatorIndex != -1) {
+        var key = Uri.decodeComponent(pair.substring(0, separatorIndex));
+        var value = Uri.decodeComponent(pair.substring(separatorIndex + 1));
 
         result[key] = value;
       }
--- a/packages/apidash_core/test/import_export/har_io_test.dart
+++ b/packages/apidash_core/test/import_export/har_io_test.dart
@@ -1,0 +1,53 @@
+import 'package:test/test.dart';
+import 'package:apidash_core/apidash_core.dart';
+
+void main() {
+  group('HarParserIO parseFormData', () {
+    late HarParserIO harParser;
+
+    setUp(() {
+      harParser = HarParserIO();
+    });
+
+    test('parses a simple key-value pair', () {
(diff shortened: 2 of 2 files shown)


animator (approved):

animator:
Thanks for the fix.

PR #1774, opened 2026-08-23
Fix #1765: Prevent stale WebSocket connections from causing resource …

### LINKED ISSUE #1765: [Bug]: Concurrent WebSocket connections can leave stale sockets untracked
Hi, while going through the current WebSocket implementation, I noticed a potential race condition in `ConnectionManager.connect()` that I haven't seen covered by the existing WebSocket issues.

`connect()` first calls `disconnect(requestId)`, but the new `WebSocket.connect()` is awaited before the socket is inserted into `_channels`. This means that if two connection attempts for the same `requestId` happen concurrently (for example, rapid repeated Connect actions or a reconnect occurring while an earlier handshake is still pending), the second `disconnect()` may not be able to cancel the first pending connection.

If both handshakes then complete, the later connection can overwrite `_channels[requestId]` while the earlier socket may remain alive but untracked. This could potentially result in multiple active WebSocket connections for a single request, duplicate incoming events, and an orphaned socket/resource leak.

I think this would be worth testing by triggering multiple connection attempts rapidly and checking the server-side connection count.

If reproducible, one possible direction would be to associate each connection attempt with a generation/token (or track pending connection attempts) and ensure that only the latest attempt can become the active connection, while closing any stale connection that completes later.

I’d be happy to investigate and reproduce this before proposing a fix. Please let me know if this is something worth opening as a separate issue.

Amresh-01:
Closes:  #1765

What changed?

Introduced a _connectionTokens map in ConnectionManager to track unique generation IDs for every connect() attempt.
Before establishing a new WebSocket.connect, a unique Object() token is generated.
After the asynchronous connection completes, the token is verified. If the token no longer matches, the connection attempt is considered stale (e.g. the user clicked "Connect" rapidly multiple times) and the socket is immediately closed, throwing a StateError.
Also patched a failing mock in window_caption_test.dart to fix type 'Null' is not a subtype of type 'FutureOr<bool>' on the main branch.
Why was this changed? Previously, concurrent connection attempts could overwrite the active channel in _channels, leaving the older connection alive but completely untracked. This orphaned socket could lead to duplicate incoming events and memory/resource leaks.

### DIFF
--- a/lib/services/connection_manager.dart
+++ b/lib/services/connection_manager.dart
@@ -25,6 +25,8 @@ class ConnectionManager {
   /// does not expose the inner socket).
   final Map<String, WebSocket> _sockets = {};
 
+  final Map<String, Object> _connectionTokens = {};
+
   /// Whether there is an active connection for [requestId].
   bool hasConnection(String requestId) => _channels.containsKey(requestId);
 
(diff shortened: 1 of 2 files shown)


Amresh-01:
Hi @ashitaprasad and @animator please review this PR

Amresh-01:
Hi @ShashwatXD Please review this PR

ShashwatXD:
maybe @badnikhil can help in this

Amresh-01:
Hi @badnikhil please review this PR

PR #1810, opened 2026-09-17
fix: clear in-memory environment state when Clear Data runs

### LINKED ISSUE #1770: Clear Data deletes persisted environment variables but leaves them active until restart
#### Describe the bug/problem

  Clear Data has inconsistent behavior for environment variables.

  The Settings page describes Clear Data as deleting request data, but the underlying reset also removes persisted environment data. However, the
  active in-memory environment state is not cleared.

  As a result, an environment variable can still be used to send requests after Clear Data, even though it will be lost after restarting API Dash.

  #### Screenshot(s)/image(s)/Video demonstrating the bug/problem

For some reason i'm unable to attach video here but to reproduce it you can-

  1. Create and enable a Global environment variable:

     baseUrl = https://httpbin.org

  2. Create and send:

     GET {{baseUrl}}/get

  3. The request succeeds with 200 OK.
  4. Go to Settings → Clear Data and confirm.
  5. The request is removed from the Requests collection.
  6. Create/send the same request again:

     GET {{baseUrl}}/get

  7. The request still succeeds with 200 OK, because baseUrl remains active in memory.

  #### Expected behavior

  Clear Data should have one consistent scope:

  - If it is intended to clear only requests, it should not delete persisted environments, history, or Dashbot data.
  - If it is intended to clear all app data, it should immediately clear active environment variables, history, and other in-memory state as well.

  At minimum, a variable deleted by Clear Data should not remain usable until the app is restarted.

  - OS: Linux
  - Version: Ubuntu 24.04 LTS
  - Browser: N/A

  #### Flutter Doctor

  Please run the flutter doctor -v command and provide the details below:[](url)

Abhist17:
## PR Description

`clearData()` (`lib/providers/collection_providers.dart:1678`) calls `hiveHandler.clear()`, which wipes the environment Hive box on disk — but `environmentsStateNotifierProvider` (`lib/providers/environment_providers.dart`) keeps its own in-memory copy that nothing invalidated after that call. So a variable deleted by Clear Data kept resolving in outgoing requests until the app was restarted, exactly as reported in #1770.

This is intentionally narrow. The issue also raises a broader design question — should Clear Data wipe everything, or only request data and preserve environments/history/Dashbot data? A previous PR (#1780) took a side on that (preserve everything except requests) and was closed without comment. This PR doesn't relitigate that: it only guarantees the issue's stated minimum bar — "a variable deleted by Clear Data should not remain usable until the app is restarted" — by making the in-memory environment state match whatever Clear Data currently deletes from disk, whichever way that broader question eventually gets settled.

**What changed:**
- Added `EnvironmentsStateNotifier.clearEnvironments()` (`lib/providers/environment_providers.dart`), which reloads environment state from the now-cleared Hive box and resyncs `environmentSequenceProvider`, `selectedEnvironmentIdStateProvider`, and `activeEnvironmentIdStateProvider` to match.
- Calls it from `CollectionStateNotifier.clearData()` right after `hiveHandler.clear()`.

**How I verified it:**
- Added `test/providers/clear_data_environment_test.dart`, which seeds a persisted Global environment variable, loads it into the provider, runs `clearData()`, and asserts both the persisted box and the in-memory provider state are empty afterward.
- Confirmed the test fails on `main` (in-memory variable still present after `clearData()`) and passes with this change.
- Ran the full `test/providers/` suite before and after: no new failures. There are 45 pre-existing failures in that directory unrelated to this change (`js_runtime_notifier_test.dart`, `ws_collection_providers_test.dart`), present identically with and without this diff — they look like a sandbox/environment issue (JS runtime, live websocket dependency), not something this PR touches.
- Ran `flutter analyze` on the touched files: no issues.
- Could not produce a video demo — no GUI/emulator available in the environment I worked in. Happy to add one if a maintainer wants it before merge.

## Related Issues

- Closes #1770

## Video Demo

- Not available from my environment (headless, no Flutter GUI/emulator). The included test reproduces the bug and demonstrates the fix at the state level; let me know if a screen recording is still needed and I'll arrange one.

### Checklist
- [x] I am using the latest Flutter stable branch (3.47.4, stable channel)
- [x] I have run the tests (`flutter test`) and all tests are passing (aside from the 45 pre-existing, unrelated failures noted above)

## Added/updated tests?

- [x] Yes — `test/providers/clear_data_environment_test.dart`

- [x] Linux


vianmangal:
Thanks for the fix! I ran automated verification with Codex on commit `6c9f9226914a232bc64063288b93d7277e01410b` on macOS (Flutter 3.47.1 / Dart 3.13.1).

- Your regression test passes. Disabling the new `clearEnvironments()` call makes it fail because `baseUrl` remains in memory, so it catches the reported bug.
- Additional local checks passed for persisted Global + active custom environments: deleted variables immediately stop resolving in request URL substitution; the custom record is removed from persistence; sequence/selection reset to Global; active environment ID/model reset to null.
- Saving/reloading does not resurrect the variables. Repeated clearing and clearing before environment-provider initialization also passed.

The provider suite, including the two additional local verification tests, finished with **118 passed, 1 skipped, 1 failed**. With the reset call disabled, it was **116 passed, 1 skipped, 3 failed**: the two additional failures were the regression checks. The remaining failure in both runs is compilation of `ui_providers_test.dart` due to undefined `headers`/`paramsMap` and Map-vs-List type errors in unchanged `lib/screens/history/history_widgets/his_request_pane.dart`, outside this PR's diff.

Analysis of your three changed files found no errors or warnings, only two informational unnecessary-import notices on unchanged imports.

Caveats: this was automated provider/state and request-substitution verification, not a manual GUI demo or actual HTTP request. Dependency setup adjusted six lockfile dependencies for the available SDK. The disabled-reset run was a behavioral control on this PR, not a clean checkout of main. No new provider-suite failures were observed with the fix enabled.

